PR #1, opened 2018-11-21
Fix C support.

lukesneeringer:
This PR refactors message creation (as well as handling of repeated fields and maps) to support the C implementation, and to adhere to the public API.


lukesneeringer:
Since the merge base is an unstable branch, I am self-merging this.

PR #2, opened 2018-11-22
Fix map instantiation in Python 3.5.

lukesneeringer:
This corrects an issue where if Python chooses the wrong dictionary order for key and value, the C implementation refuses to accept the descriptor outright.

(Python 3.6 and above always use a consistent order, but Python 3.5 chooses the order randomly.)


lukesneeringer:
As this is targeting an unstable branch, merging outright.

PR #35, opened 2020-07-12
feat: add support for proto3 optional fields

software-dov:
Due to popular demand, proto3 recently added the concept of field presence. Fields can be marked as 'optional', which is internally represented via a synthetic _oneof.
See https://github.com/protocolbuffers/protobuf/releases/tag/v3.12.0 for additional background and information.

Generated messages can define optional fields, and message classes gain constants that can be used to check for field presence in instances. These constants are not visible to message instances and therefore don't collide with message field names.

Fields can be set as 'optional' using the following syntax

Example:
```python
class Squid(proto.Message):
    mantle = proto.Field(proto.INT32, number=1, optional=True)

s = Squid()
assert not Squid.mantle in s
s.mantle = 2112
assert Squid.mantle in s
```

### DIFF
--- a/proto/message.py
+++ b/proto/message.py
@@ -160,12 +160,29 @@ class MessageMeta(type):
 
         # As per descriptor.proto, all synthetic oneofs must be ordered after
         # 'real' oneofs.
+        opt_attrs = []
         for field in fields:
             if field.optional:
                 field.oneof = "_{}".format(field.name)
                 field.descriptor.oneof_index = oneofs[field.oneof] = len(oneofs)
-                fget = lambda self: self.__class__.pb(self).HasField(field.name)
-                attrs['Has_{}'.format(field.name)] = property(fget, None)
+                opt_attrs.append(field.name)
+
+        # Generating a metaclass dynamically provides class attributes that
+        # instances can't see. This provides idiomatically named constants
+        # that provide the following pattern to check for field presence:
+        #
+        # class MyMessage(proto.Message):
+        #     field = proto.Field(proto.INT32, number=1, optional=True)
+        #
+        # m = MyMessage()
+        # MyMessage.field in m
+        class AttrsMeta(mcls):
+            pass
+
+        for a in opt_attrs:
+            setattr(mcls, a, a)
+
+        mcls = AttrsMeta
 
         # Determine the filename.
         # We determine an appropriate proto filename based on the
--- a/proto/message.py
+++ b/proto/message.py
@@ -129,7 +129,6 @@ class MessageMeta(type):
 
             # Add the field to the list of fields.
             fields.append(field)
-
             # If this field is part of a "oneof", ensure the oneof itself
             # is represented.
             if field.oneof:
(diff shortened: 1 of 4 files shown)


lukesneeringer:
Why `s.Has_mantle` and not something more idiomatic, such as (my order of preference):

- `'mantle' in s`
- `hasattr(s, 'mantle')`
- `Squid.has_mantle(s)`

I get that `s.has_mantle` is a potential collision issue but there are alternatives that follow PEP 8.

proto/message.py
```diff
@@ -159,6 +158,15 @@ def __new__(mcls, name, bases, attrs):
...
             index += 1
 
+        # As per descriptor.proto, all synthetic oneofs must be ordered after
+        # 'real' oneofs.
+        for field in fields:
+            if field.optional:
+                field.oneof = "_{}".format(field.name)
+                field.descriptor.oneof_index = oneofs[field.oneof] = len(oneofs)
+                fget = lambda self: self.__class__.pb(self).HasField(field.name)
+                attrs['Has_{}'.format(field.name)] = property(fget, None)
```
lukesneeringer, line 168: It seems like logic similar to this could be added to `__contains__`, which I think is the right thing to do.

proto/message.py
```diff
@@ -374,8 +382,8 @@ def __init__(self, mapping=None, **kwargs):
 
     def __bool__(self):
         """Return True if any field is truthy, False otherwise."""
-        return any([k in self and getattr(self, k)
-                    for k in self._meta.fields.keys()])
+        return any(k in self and getattr(self, k)
+                   for k in self._meta.fields.keys())
 
     def __contains__(self, key):
         """Return True if this field was set to something non-zero on the wire.
```
lukesneeringer, line 389: And, in fact, this is precisely how `__contains__` is advertised here.

lukesneeringer: Actually, reading the code, I think it even _already works_.

software-dov:
I had a few design goals that led to this implementation:
1) A similar interface to the underlying mechanism, i.e. `HasField`
2) Something that minimized the risk of name collision.
3) To the greatest extent possible, no user provided strings.

I agree, the generated naming convention is less than ideal, and using `__contains__` is better, but I'd still like to avoid making the user pass in strings. I've got an idea on how to deal with that.

lukesneeringer:
I would add (4) PEP 8 compliance to that list, which is my complaint here. In particular, I dislike it when we force PEP 8 _non_-compliance into user code, which protobuf has a bad habit of doing.

software-dov:
Although adding `Squid.has_mantle` via a metaclass instance is the lowest on your order of preference, what do you think about combining that with `__contains__`? The field names become class only attributes that are used as constants to prevent the user from having to pass in strings manually.

Something like ```python
s = Squid(mantle=10)
Squid.mantle in s
>>> True
s.mantle
>>> 10
s = Squid()
Squid.mantle in s
>>> False
s.mantle
>>> 0
```

proto/message.py
software-dov: Changed the interface; this code no longer exists.

proto/message.py
software-dov: Yup. Turns out I reinvented a wheel :P

lukesneeringer:
That is a clever idea (and borrows from SQL Alchemy's concept of fields).

I think that is fine, but can I make a YAGNI argument? Essentially, since `__contains__` already works, stick with that, and if there is user confusion, then add this (or one of the other potential options)?

software-dov:
Ads wants to use optional fields and test for presence. This feature is explicitly at their request, and they wanted an ergonomic mechanism to detect presence. Since this is a reactive and not a speculative feature, I'd tentatively say YAGNI doesn't apply.

lukesneeringer:
I assert that mirroring `dict.__contains__` is ergonomic though. :) That said, as long as `__contains__` _works_, I am not too concerned.

PR #36, opened 2020-07-13
Bump version to 2.0.0

software-dov:


busunkim96:
Is proto3 presence a breaking change? If it is just a new feature, I think we can just bump the minor (1.0.0 -> 1.1.0)

software-dov:
I was originally taking a conservative view on that point since the dependency boost from protobuf 3.5.1 to 3.12.0 is large, and I didn't want users to be surprised.
On second thought, given that the proto-plus interface is backwards compatible, and the protobuf dependency bump is a minor version change, I feel more sanguine about making this a minor version release.

lukesneeringer:
I think a minor version change is fine.

Issue #38, opened 2020-07-14
Add support for DatetimeWithNanos on pb Timestamp


crwilcox:
In Google Cloud API Core, we have a class, `DatetimeWithNanoseconds`. This allows us to keep `timestamp_pb2` precision, but have pythonic interfaces.

https://github.com/googleapis/python-api-core/commit/5e121ce063bddfee174cff603668c5191876fd08

It would be great if protoplus could implement this.

lukesneeringer:
My vote on how to do this would be to move `DatetimeWithNanoseconds` from api-core to here, and then have api-core import it and put it in `__all__` in the module where it used to be.

crwilcox:
Funny you would comment, I was working on this now. just PR'd at https://github.com/googleapis/proto-plus-python/pull/40

PR #40, opened 2020-07-16
feat: add DatetimeWithNanoseconds class to maintain Timestamp pb precision.

crwilcox:
Ports pieces of https://github.com/googleapis/python-api-core/blob/5e5559202891f7e5b6c22c2cbc549e1ec26eb857/google/api_core/datetime_helpers.py to protoplus. Particularly, a datetime implementation that provides nanoseconds.

Fixes https://github.com/googleapis/proto-plus-python/pull/38


docs/marshal.rst
```diff
@@ -13,24 +13,24 @@ Well-known types
...
+``google.protobuf.BytesValue``      ``bytes``                                          Yes
+``google.protobuf.DoubleValue``     ``float``                                          Yes
+``google.protobuf.Duration``        ``datetime.timedelta``                               –
+``google.protobuf.FloatValue``      ``float``                                          Yes
+``google.protobuf.Int32Value``      ``int``                                            Yes
+``google.protobuf.Int64Value``      ``int``                                            Yes
+``google.protobuf.ListValue``       ``MutableSequence``                                  –
+``google.protobuf.StringValue``     ``str``                                            Yes
+``google.protobuf.Struct``          ``MutableMapping``                                  –
+``google.protobuf.Timestamp``       ``proto.datetime_helpers.DatetimeWithNanoseconds`` Yes
```
lukesneeringer, line 29: This arguably obscures more than it elucidates. It might be more valuable to put `datetime.datetime` with a footnote that says, "Actually, a subclass with nanosecond precision". This is consistent with the use of `MutableSequence` and `MutableMapping`, which are actually interfaces.

What 99% of people want to know is that they can use it like a datetime.

proto/datetime_helpers.py
```diff
@@ -0,0 +1,219 @@
...
+"""Helpers for :mod:`datetime`."""
+
+import calendar
+import datetime
+import re
+
+from google.protobuf import timestamp_pb2
+
+
+_UTC_EPOCH = datetime.datetime.utcfromtimestamp(0).replace(tzinfo=datetime.timezone.utc)
```
lukesneeringer, line 24: Nit: This seems to be more than 79 characters.

tests/test_datetime_helpers.py
```diff
@@ -0,0 +1,294 @@
...
+def test_to_rfc3339_with_non_utc_ignore_zone():
+    zone = pytz.FixedOffset(-60)
+    value = datetime.datetime(2016, 4, 5, 13, 30, 0, tzinfo=zone)
+    expected = "2016-04-05T13:30:00.000000Z"
+    assert datetime_helpers._to_rfc3339(value, ignore_zone=True) == expected
+
+
+class Test_DateTimeWithNanos(object):
+    @staticmethod
+    def test_ctor_wo_nanos():
```
software-dov, line 67: Why have a class that wraps the test methods, and an old-style class at that? Is this something clever I don't understand or cruft?

crwilcox: There are two styles. I lifted these as is. I can transform though. I can also put it in and make an issue to reformat?  As written though, only one test case is new, to cover me implementing replace for the derived datetime class.

crwilcox: altered.

PR #54, opened 2020-07-17
Tweak pypi circleci task

software-dov:


busunkim96:
Blargh, I forgot that adding this repo to sloth would have the side effect of it adding the required checks from the other Python repos.  I just manually removed the Kokoro check, but we'll have to figure out something more permanent.

busunkim96:
Opened https://github.com/googleapis/repo-automation-bots/pull/709/ to adjust the required checks.

PR #61, opened 2020-07-22
feat: support fixed filename salt to allow proto-plus use with schema registry tools

joncourt:
By default the underlying protocol buffer class is created with an 8 character random salt. This allows re-use of the filename by other proto messages if needed (e.g. if __all__ is not used).

This means new filenames (and hence proto schema names) are created each time the underlying proto classes are generated which is at startup i believe. Normally this is not visible and causes no issue however if you are using a schema registry this presents a problem. With a schema registry the name of the underlying schema becomes important and must remain consistent if the schema registry is to perform its function.

This PR changes the salting mechanism to use the class name instead so making it deterministic and predictable.


lukesneeringer:
Are there any risks to just using the "append the message name" variant all the time? I might prefer just change the behavior over making an option.

noxfile.py
```diff
@@ -18,7 +18,7 @@
 import nox
 
 
-@nox.session(python=["3.5", "3.6", "3.7"])
+@nox.session(python=["3.5", "3.6", "3.7", "3.8"])
```
lukesneeringer, line 21: Thanks. :-)

joncourt:
I agree with just changing the mechanism if you're ok with that and don't believe it will break anything. I prefer it actually. 

As far as MD5 is concerned it makes sense technically and i agree. The names become visible in a schema reg so if there is a way to make it nicer I'd prefer it though. I think from your comment that I've missed something about the naming that I've done so i'm going to give it a bit more study and get back to you. This is a first foray into Python for an old java dev so look out for that 😁. 

In the meantime I'll unwind the 'making it an option' bit and update the underlying mechanism wholesale.

joncourt:
This project is a life saver btw. Was really worried about convincing the devs here to accept a code generation step 😄. Thanks for putting it together.

joncourt:
> If `__all__` _is_ used (which is most of the time since our code generator uses it), it might seem a bit weird to have a bunch of messages and have the _last_ message's name be appended.

I'm don't understand what you say here. If there are multiple declarations in a .py file they come out as one .proto file each - hence the issue with duplication. It seems reasonable to add the classname (or message name depending on perspective). Are you saying in the imports and usage you'll see the class/message name twice?

proto/_file_info.py
```diff
@@ -47,8 +47,9 @@ def generate_file_pb(self):
         # Salt the filename in the descriptor.
         # This allows re-use of the filename by other proto messages if
         # needed (e.g. if __all__ is not used).
+        salt = "" if salt is None else salt
```
software-dov, line 50: Nit: can do
```python
salt = salt or ""
```

proto/message.py
```diff
@@ -580,4 +581,4 @@ def pb(self) -> Type[message.Message]:
         return self._pb
 
 
-__all__ = ("Message",)
```
software-dov, line 583: `__all__` should still be a collection.
```python
__all__ = ["Message",]
```

lukesneeringer:
> I don't think I understand what you say here. If there are multiple declarations in a .py file they come out as one .proto file each - hence the issue with duplication. It seems reasonable to add the classname (or message name depending on perspective). Are you saying in the imports and usage you'll see the class/message name twice?

Not quite true. It does that _if it does not have a manifest of what is in the file_. The code that checks that is [here](https://github.com/googleapis/proto-plus-python/blob/master/proto/_file_info.py#L119-L129).
You can provide a manifest:

```python
__protobuf__ = proto.module(
    package='foo.bar.v1',
    manifest={
        'Foo',
        'Bar',
    },
)
class Foo(proto.Message):
    # ...

class Bar(proto.Message):
    # ...
```

Given this code, the first thing that happens is that the `Foo` message is created. When it goes to generate the file, it checks the manifest, realizes that `Bar` is coming, and holds off.

proto/message.py
lukesneeringer: Specifically, it should be a tuple.

joncourt:
Ahh. Nice.

Ok I'll take another look.

joncourt:
Thanks Dov. I'll fix it up.

software-dov:
If you don't mind my asking, can you tell us a little about the project? Your mention of code generation has us intrigued.

joncourt:
> If you don't mind my asking, can you tell us a little about the project? Your mention of code generation has us intrigued.

I don't mind at all.

We have a large pipeline we process through many steps including several machine learning steps. All of this is supported using Kafka and it's all a little loose. It's JSON based and parsed simply using dict in python tooling. Currently we have many topics with several terabytes of data - most of it duplicated in other topics.

As a part of improving our data volume, performance metrics and code robustness we're looking at introducing the Kafka Schema Registry, moving to a binary format and using a more structured approach to dealing with the data we're working with from kafka. This means using well defined class structures to represent our data, a schema'd data format and managed data contracts between our services. Another advantage to moving to a new protocol is that it forces a clean break from our current data style and prevents our current style leaking too much as we role through our pipeline over the next year or two.

As for code generation when using Protobuf in particular there is a necessary code generation step with protoc to translate the proto format to a format usable in python, and if we use the mypy plugin we get some decent pyi stub files to help with type info in our development environment. Being an old java dev i don't mind code generation at all because it's generally faster and safer than reflection, but i'm in a python shop now and it's a much harder sell it seems.

proto-plus solves all of these problems neatly by allowing declaration directly in python and avoiding code generation. Additionally it uses with a structure that makes it clear that this is not just any old class but one that will be a serialization format for RPC or messaging. I think this is important because it helps prevent us falling into the old trap of hiding RPC or messaging behind a local facade and ignoring all the extra problems you encounter with remote calls and shared data.

Using Avro is an option which can be used as a dict but the schema format is annoying and ugly to declare, and it's notoriously slow in Python for serialization. Plus i like the idea of a class structure behind our data processing (that might just be the java dev in me though). JSON Schema is an option but isn't such a clean break plus it's still json with all it's verbosity, volume and slow serialization (ignoring compression on the brokers). On top of this is the general support of Protobuf - it's everywhere and it opens the door to using gRPC in our API calls - particularly given we've got ML models in services which may require a lot of data in any particular call.

In the end it may actually be not such a problem generating because we'll package up shared models into libs so the generation is hidden away. We also use Go in some places and we expect to use the Schema Registry to share schemas with those services - if it doesn't work as well as we like though then we'll be back to declaring plan .proto schemas there.

joncourt:
I've applied your suggestion @lukesneeringer. It now checks for a manifest and applies a classname salt if there isn't one. It also logs a warning if there is a manifest but a class isn't mentioned in it. Not sure if this is a good idea - generally if it would remind me to mention the class in the manifest if i've forgotten though.

joncourt:
in case you're interested and perhaps able to help... this is what integration with the Schema Reg looks like (with Faust).

I don't fully understand why for the todo's you see in there.

`from typing import Any

from confluent_kafka.schema_registry import SchemaRegistryClient
from confluent_kafka.schema_registry.protobuf import (
    ProtobufSerializer,
    ProtobufDeserializer,
)
from confluent_kafka.serialization import MessageField, SerializationContext
from faust.serializers import codecs


class RegisteredProtoCodec(codecs.Codec):
    def __init__(
        self,
        topic_name: str,
        msg_type: Any,
        schema_registry_client: SchemaRegistryClient,
        is_key: bool = False,
        **kwargs: Any
    ) -> None:
        super().__init__(children=(), **kwargs)
        self.msg_type = msg_type
        self.context = SerializationContext(
            topic_name, MessageField.KEY if is_key else MessageField.VALUE
        )
        self.serializer = ProtobufSerializer(msg_type, schema_registry_client)
        self.deserializer = ProtobufDeserializer(msg_type)

    def _dumps(self, obj: Any) -> bytes:
        return self.serializer(obj, self.context)

    def _loads(self, s: bytes) -> Any:
        return self.deserializer(s, self.context)


class RegisteredProtoPlusCodec(codecs.Codec):
    def __init__(
        self,
        topic_name: str,
        msg_type: Any,
        schema_registry_client: SchemaRegistryClient,
        is_key: bool = False,
        **kwargs: Any
    ) -> None:
        super().__init__(children=(), **kwargs)
        self.msg_type = msg_type
        self.context = SerializationContext(
            topic_name, MessageField.KEY if is_key else MessageField.VALUE
        )
        self.serializer = ProtobufSerializer(msg_type.pb(), schema_registry_client)
        self.deserializer = ProtobufDeserializer(msg_type.pb())

    def _dumps(self, obj: Any) -> bytes:
        # Nested proto-plus objects don't seem to be fully wrapped. Testing for `_pb` attr distinguishes
        # fully wrapped from not.
        if hasattr(obj, "_pb"):
            return self.serializer(type(obj).pb(obj, coerce=True), self.context)
        else:
            # TODO: Recursing down a class structure is meant to unwrap the protobuf class from the proto-plus wrapper on the fly - this
            #       doesn't always seem to get the __class__ correct (seems to drop the namespace). This hack forces the __class__
            #       back to the intended so instanceof(..) operations work correctly.
            obj.__class__ = self.msg_type._meta.pb().__class__
            return self.serializer(obj, self.context)

    def _loads(self, s: bytes) -> Any:
        pb = self.deserializer(s, self.context)
        # TODO: The deserialization seems to drop the namespace of the class name meaning isinstance(before, after) is always false
        #       when engaging the confluent schema-regsitry serializer/deserializer (official name: serdes). A hack is to force the
        #       __class__ name post deserialization and before re-wrapping into the proto-plus version so instanceof(..) operations
        #       work correctly.
        pb.__class__ = self.msg_type._meta.pb().__class__
        result = self.msg_type.wrap(pb)
        return result
`

joncourt:
great. thanks guys.

PR #63, opened 2020-07-22
feat: prevent unnecessary copies when deserializing proto

software-dov:
Prevent unnecessary data copies when deserializing proto-plus messages.

### DIFF
--- a/proto/message.py
+++ b/proto/message.py
@@ -345,7 +345,9 @@ class MessageMeta(type):
             ~.Message: An instance of the message class against which this
             method was called.
         """
-        return cls(Parse(payload, cls()._pb))
+        instance = cls()
+        Parse(payload, instance._pb)
+        return instance
 
 
 class Message(metaclass=MessageMeta):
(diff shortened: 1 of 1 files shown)


proto/message.py
```diff
@@ -341,7 +345,9 @@ def from_json(cls, payload) -> "Message":
             ~.Message: An instance of the message class against which this
             method was called.
         """
-        return cls(Parse(payload, cls()._pb))
+        instance = cls()
+        Parse(payload, instance._pb)
+        return instance
```
busunkim96, line 350: I'm having trouble wrapping my head around this - does calling `Parse` modify the instance that's passed in?

busunkim96: Never mind! I read the [docs](https://googleapis.dev/python/protobuf/latest/google/protobuf/json_format.html#google.protobuf.json_format.Parse) on this and it makes sense now 🙃

busunkim96:
@software-dov FYI, I think the PyPI push is still borked. https://app.circleci.com/jobs/github/googleapis/proto-plus-python/1175 I published 1.3.1 and 1.3.2 manually

PR #99, opened 2020-08-10
Repeated insertion

bobhancock:
The refactor will pass unit test but will only work if the sequence is a Python list, which it is not guaranteed to be.


bobhancock:
Incorrect revert.

PR #107, opened 2020-08-27
feat: allow enum strings in json serialization and deserialization

software-dov:
For protobuf messages that contain enum fields, it is now possible to
specify that enum variants should be serialized as names and not as integers.

E.g.

json_str = MyMessage.to_json(my_message, enum_strings=True)

Similarly, serialization from json that uses this convention is now supported.

This is useful for interoperation with other data sources that do use
strings to define enum variants in json serialization; and for
debugging, where visually inspecting data structures can be helpful,
and variant names are more informative than numerical values.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -249,7 +249,7 @@ jobs:
       - run:
           name: Format files
           command: |
-            black .
+            black -l 88 .
       - run:
           name: Check diff
           command: |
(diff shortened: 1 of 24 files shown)


software-dov:
Please ignore the style check failure: [Black](https://github.com/psf/black) got updated recently, and it screwed with various wrappings that were previously okay.

Once I have approval, I can either re-run black on the the entire code base, or I can punt that into a subsequent review.

software-dov:
In general, I'm not inclined to recreate and pass through all the json serialization options. I agree, changing the behavior to be stringy by default would be a breaking change at this point.

Issue #103, opened 2020-08-17
marshalling nested enums from python strings raises a TypeError


software-dov:
This requires marshaling from python structures.

Minimal reproduction:
```python
import proto

class Zone(proto.Enum):
    LITTORAL = 1
    MESOPELAGIC = 2
    ABYSSAL = 3
    HADEAN = 4

class Squid(proto.Message):
    zone = proto.Field(Zone, number=1)

class Trawl(proto.Message):
    squids = proto.RepeatedField(Squid, number=1)   # Note: this indirection with the nested field
                                                    # is necessary to trigger the exception.
                                                    # Setting the field in an existing message accepts strings AND
                                                    # checks for valid variants.
                                                    # Similarly, constructing a message directly with a top level 
                                                    # enum field kwarg passed as a string is also handled correctly, i.e.
                                                    # s = Squid(zone="ABYSSAL")
                                                    # does NOT raise an exception.

t = Trawl(squids=[{"zone": "HADEAN"}])
```

generates

```
TypeError: 'HADEAN' has type str, but expected one of: int, long
```

This is related to #52, and most of the implementation will probably be shared.

software-dov:
A workaround is to not use strings directly but instead to use Enum class attributes.
E.g., assuming the above example class definitions,
```python
t = Trawl(squids=[{"zone": Zone.HADEAN}, {"zone": Zone.LITTORAL}])
```

hkdevandla:
Making this as p3 considering that there is a workaround. @software-dov , adjust this if you disagree. thanks!

software-dov:
There is a workaround, and the issue is _fixed_ by #107

software-dov:
Fixe

PR #118, opened 2020-09-04
fix: third party enums don't break first class enums

software-dov:
For real, this time.
Third party, unwrapped enums no longer cause problems
if they are used to describe proto-plus fields.

Limitations:
* Third party enums cannot be passed by string name to a Field
* More scenarios require a module definition than previously
** Defining a proto.Enum may cause a FileDescriptor to be added to the
   descriptor pool. The only way to defer until _all_ Enums and
   Messages in a module have been added is to provide a manifest or
   make the Enum a nested type of a Message.

   This mostly affects unit tests and interactive development.

Includes re-addition of stringy enums in JSON.
Includes re-addition of fix for #103

Release-As: 1.10.0-dev1

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -24,17 +24,22 @@ def unit(session, proto="python"):
 
     session.env["PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION"] = proto
     session.install("coverage", "pytest", "pytest-cov", "pytz")
-    session.install("-e", ".")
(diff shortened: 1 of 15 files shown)


software-dov:
Python-dlp passes all unit tests using both the python and cpp implementations.

lukesneeringer (approved):

busunkim96 (approved):

PR #123, opened 2020-09-10
fix: loosen tag regex for publish_package

busunkim96:
Release-As: 1.10.0-dev2


busunkim96:
Arbitrarily bumping to `dev2` since that is easier than fighting with the existing tag

software-dov (approved):

PR #144, opened 2020-10-08
Fix license copyright

marcelotrevisani:


marcelotrevisani:
@googlebot I signed it!

marcelotrevisani:
I faced this problem with the license when I was creating a conda package for ``proto-plus``

tswast:
Per https://opensource.google/docs/releasing/preparing/#license-headers

> For the Apache License, simply copy the official 2.0 license into your LICENSE file **verbatim** (do not edit the Copyright section in the template, only when you use that template in the header of your source code files.)

Issue #141, opened 2020-10-05
LICENSE file not included in source distribution on PyPI


tswast:
It's a best practice to include a LICENSE file in the source distribution. This way licenses can be audited by automated tooling.

Also, this is recommended to pass the linter on conda-forge.

busunkim96:
@software-dov Unless you have any objections I'll open a PR to add a `MANIFEST.in`. It will also take care of #143. 

https://packaging.python.org/guides/using-manifest-in/

busunkim96:
Closed by #146

PR #146, opened 2020-10-08
fix: add LICENSE and tests to package

### LINKED ISSUE #143: Include tests with PyPi packages?
The PyPi packages are missing the tests directory, unless this is intentional could they be included?

busunkim96:
Closes #143 and #141 

Adds [MANIFEST.in](https://packaging.python.org/guides/using-manifest-in/) so the license and tests are included in the published package.


tswast (approved):
Thank you!

PR #164, opened 2020-11-20
feat: add default values parameter to to_json

yon-mg:


tests/test_json.py
```diff
@@ -97,6 +97,29 @@ class Zone(proto.Enum):
...
+    json1 = (
+        Squid.to_json(s, including_default_value_fields=False)
+        .replace(" ", "")
+        .replace("\n", "")
+    )
+    assert json1 == '{"name":"Steve"}'
+
+    json2 = Squid.to_json(s).replace(" ", "").replace("\n", "")
+    assert (
+        json2 == '{"name":"Steve","massKg":0}' or json2 == '{"massKg":0,"name":"Steve"}'
```
software-dov, line 115: Is the ordering not stable?

software-dov (approved):
LGTM.

tests/test_json.py
yon-mg: No but it shouldn't matter for json objects right?

software-dov: I don't think it should matter, I just was curious if you were running into problems during testing.

proto/message.py
```diff
@@ -327,7 +327,13 @@ def deserialize(cls, payload: bytes) -> "Message":
         """
         return cls.wrap(cls.pb().FromString(payload))
 
-    def to_json(cls, instance, *, use_integers_for_enums=True) -> str:
+    def to_json(
+        cls,
+        instance,
+        *,
+        use_integers_for_enums=True,
+        including_default_value_fields=True
```
busunkim96, line 335: @yon-mg Could you open a follow up PR to update the docstring below?

This will result in a change in behavior for someone already using this method Protobuf defaults to `False` for this option. https://googleapis.dev/python/protobuf/latest/google/protobuf/json_format.html#google.protobuf.json_format.MessageToJson

yon-mg: I'm not sure I understand. I did not change the way `google.protobuf.json_format.MessageToJson` works. In fact this method `message.MessageMeta.to_json` should call `MessageToJson` in the same way as it did before if no changes are made to the additional parameter I added.

yon-mg: Although I did neglect to update the docstring for this method which I will update shortly if that's what you mean.

PR #189, opened 2021-02-18
feat: Pypi publish ghub actions

software-dov:


busunkim96 (approved):
Not sure if it's already the case, but if possible please use an [API token](https://pypi.org/help/#apitoken)

software-dov:
I have the environment setup with an API token. Whether it works is another issue :p

software-dov:
@0x2b3bfa0 and here too please :P
This is totally my fault. Thank you for contributing to the project! It's so nice to see that other people use it.

0x2b3bfa0:
@googlebot I consent.

PR #215, opened 2021-03-15
feat: add copy_from method for field assignment

### LINKED ISSUE #190: `=` should have transitive assignment for nested fields, not copying behavior
The `=` operator is recommended to use in place of `CopyFrom` for nested message fields, so for example:

```py
# with proto-plus
message.nested = copy

# with raw protobuf messages
message.nested.CopyFrom(copy)
```

Syntactically this is nice, however it introduces unexpected behavior because the `=` operation in the first statement implicitly makes a copy of the second message.

With other Python objects, and in other programming languages, that `=` operator would make a transitive assignment so that `message.nested` and `copy` point to the same object in memory. So for example when using a `dict`:

```py
message = {}
copy = {"foo": "bar"}
message["copy"] = copy
message["copy"]["another"] = 1
print(copy)
>>> {"foo": "bar", "another": 1}
```

It might be better for the `=` operator to behave as it would for a `dict` or other objects, and to either expose a new interface specifically for copying objects, or to rely on conventional Python patterns like using the [`copy`](https://docs.python.org/3/library/copy.html) module.

software-dov:
Because of implementation detatils of the underlying protocol buffers
runtime, assigning to a proto-plus message field is achieved by
copying, not by updating references. This can lead to surprising and
unintuitive behavior for developers who are expecting python object
behavior, e.g. reference aliases.

This PR adds a 'Message.copy_from' method that is semantically
identical to regular assignment. This can be used at the discretion of
the developer to clarify expected behavior.

### DIFF
--- a/docs/messages.rst
+++ b/docs/messages.rst
@@ -89,6 +89,7 @@ objects as possible. It is possible to update a message's field by assigning to 
 just as if it were a regular python object.
 
 .. code-block:: python
+
    song = Song()
    song.composer = Composer(given_name="Johann", family_name="Bach")
 
@@ -141,6 +142,7 @@ just as if it were a regular python object.
       # We can use 'copy_from' if we're concerned that the code
       # implies that assignment involves references.
       composer = Composer(given_name="Elisabeth", family_name="Bach")
+      # We could also do Message.copy_from(song.composer, composer) instead.
       Composer.copy_from(song.composer, composer)
 
       assert song.composer.given_name == "Elisabeth"
--- a/docs/messages.rst
+++ b/docs/messages.rst
@@ -113,7 +113,7 @@ just as if it were a regular python object.
 
    :class:`proto.Message` defines a helper message, :meth:`~.Message.copy_from` to
    help make the distinction clear when reading code.
-   The semantics of :meth:`~.Message.copy_from` are addentical to regular assignment.
+   The semantics of :meth:`~.Message.copy_from` are identical to the field assignment behavior described above.
 
    .. code-block:: python
 
--- a/docs/messages.rst
+++ b/docs/messages.rst
@@ -80,6 +80,72 @@ Instantiate messages using either keyword arguments or a :class:`dict`
     >>> song.title
     'Canon in D'
 
+
+Assigning to Fields
+-------------------
+
+One of the goals of proto-plus is to make protobufs feel as much like regular python
+objects as possible. It is possible to update a message's field by assigning to it,
+just as if it were a regular python object.
+
+.. code-block:: python
+   song = Song()
+   song.composer = Composer(given_name="Johann", family_name="Bach")
+
+   # Can also assign from a dictionary as a convenience.
+   song.composer = {"given_name": "Claude", "family_name": "Debussy"}
+
+   # Repeated fields can also be assigned
+   class Album(proto.Message):
+       songs = proto.RepeatedField(Song, number=1)
+
+   a = Album()
+   songs = [Song(title="Canon in D"), Song(title="Little Fugue")]
+   a.songs = songs
+
+.. note::
+
+   Assigning to a proto-plus message field works by making copies, not by updating references.
+   This is necessary because of memory layout requirements of protocol buffers.
+   These memory constraints are maintained by the protocol buffers runtime.
+   This behavior can be surprising under certain circumstances, e.g. trying to save
+   an alias to a nested field.
+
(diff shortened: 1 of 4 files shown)


arithmetic1728 (approved):

docs/messages.rst
```diff
@@ -80,6 +80,72 @@ Instantiate messages using either keyword arguments or a :class:`dict`
...
+
+   Assigning to a proto-plus message field works by making copies, not by updating references.
+   This is necessary because of memory layout requirements of protocol buffers.
+   These memory constraints are maintained by the protocol buffers runtime.
+   This behavior can be surprising under certain circumstances, e.g. trying to save
+   an alias to a nested field.
+
+   :class:`proto.Message` defines a helper message, :meth:`~.Message.copy_from` to
+   help make the distinction clear when reading code.
+   The semantics of :meth:`~.Message.copy_from` are addentical to regular assignment.
```
BenRKarl, line 116: `addentical` should be `identical` 

And perhaps `... identical to the field assignment behavior described above.` would help clarify, as "regular assignment" could imply the behavior most would expect outside the context of protobufs.

proto/message.py
```diff
@@ -602,6 +602,35 @@ def __setattr__(self, key, value):
...
+    def copy_from(self, other):
+        """Equivalent for protobuf.Message.CopyFrom
+
+        Args:
+            other: (Union[dict, ~.Message):
+                A dictionary or message to reinitialize the values for this message.
+        """
+        if isinstance(other, type(self)):
+            # Just want the underlying proto.
+            other = Message.pb(other)
```
BenRKarl, line 614: This is equivalent to doing `other = other._pb` right?

proto/message.py
```diff
@@ -602,6 +602,35 @@ def __setattr__(self, key, value):
...
+        elif isinstance(other, type(Message.pb(self))):
```
BenRKarl, line 615: Why does it `pass` in this scenario?

proto/message.py
```diff
@@ -602,6 +602,35 @@ def __setattr__(self, key, value):
...
+            # Don't need to do anything.
+            pass
+        elif isinstance(other, collections.abc.Mapping):
```
BenRKarl, line 618: Is this is in case of a repeated field?

tests/test_message.py
```diff
@@ -317,3 +317,27 @@ class Squid(proto.Message):
...
+def test_copy_from():
+    class Mollusc(proto.Message):
+        class Squid(proto.Message):
+            mass_kg = proto.Field(proto.INT32, number=1)
+
+        squid = proto.Field(Squid, number=1)
+
+    m = Mollusc()
+    s = Mollusc.Squid(mass_kg=20)
+    Mollusc.Squid.copy_from(m.squid, s)
```
BenRKarl, line 331: The reason we can't have `s.squid.copy_from` is because of potential collisions with the proto definition, right?

docs/messages.rst
software-dov: Doh! Good catch. Yes, the proposed wording is more clear.

proto/message.py
software-dov: Yes, it is equivalent. The code (and documentation) tries to use the accessor instead of touching the attribute directly.

proto/message.py
software-dov: Because there's no transformation that needs to occur. It's just here to catch the input type and make sure that it doesn't generate the type mismatch exception below.

proto/message.py
software-dov: It's to catch the dictionary syntax:
```python
Message.copy_from(message.submessage, {"name": "Lawrence", "armor_class": 15})
```

tests/test_message.py
software-dov: Correct. The general problem is that we have to be very, very careful about adding fields and methods to the message class types because of potential name collisions with fields, and the general solution is to add the method (or field) to the metaclass. This means that method invocation becomes a little more indirect: we invoke it from the _class_ and pass in the instance as the first parameter.

BenRKarl: 👍🏻

proto/message.py
BenRKarl: Oh good to know that's possible!

proto/message.py
BenRKarl: 👍🏻

proto/message.py
BenRKarl: 👍🏻

BenRKarl (approved):

PR #217, opened 2021-03-16
feat: Add support for forthcoming `google.type.Decimal`.

lukesneeringer:
This PR converts between the (not yet released) `google.type.Decimal` and Python's `decimal.Decimal`.

THE TESTS WILL FAIL RIGHT NOW because the dependent type is unreleased. Once Decimal is released, we should ensure this works and only then can we merge it.


software-dov:
I cheated with a locally produced mockup of Decimal and tweaked/augmented tests as a result. It should pass when the real thing is merged.

software-dov (approved):

busunkim96:
Is there a issue/PR to track for when this arrives in protobuf?

parthea:
This change requires adding a dependency on `googleapis-common-protos`. I'm going to hold off on approving/merging this change until there is a use case to avoid taking on another dependency. If this change is required, please open a new issue via https://github.com/googleapis/proto-plus-python/issues/new/choose.

PR #226, opened 2021-04-13
fix: lookup attribute instead of performing a deepcopy

### LINKED ISSUE #224: Very poor performance in Firestore reads due to proto-plus sequence unmarshaling
### Environment details
- OS type and version: Windows 10 Pro
- Python version: 3.9
- pip version: 20.2.2
- google-cloud-firestore: 2.1.0

### Problem
We have been using Firestore with the python admin SDK for some time now, and in some cases the latency in fetching data was nowhere near the expected latency. For example, reading 500 (very small) documents from a collection could take up to 30 
seconds depending on which fields the documents had. After some experiments, we found out this poor performance directly related with the presence of empty arrays in the documents. 

### Seemingly related issues
- #222

### Steps to reproduce
- Create two collections of 500 documents each: one with documents containing an empty array, and the other collection with documents containing a non-empty array.
- Retrieve all documents in each collection and compare execution time.
```python
from google.cloud import firestore
import time
import os

N_DOCS = 500

if __name__ == '__main__':

    os.environ['GOOGLE_APPLICATION_CREDENTIALS'] = "./gcp_sa_creds-dev.json"
    firestore_client = firestore.Client()

    for coll_name in ["performance_test_empty_array", "performance_test_not_empty_array"]:
        coll_ref = firestore_client.collection(coll_name)

        print(f"\n{coll_name}")
        
        # Just uncomment this if you need to clean up the previous test
        # batch = firestore_client.batch()
        # for doc in coll_ref.stream():
        #     batch.delete(doc.reference)
        # batch.commit()

        # Create data
        print("Creating data")
        start = time.time()
        batch = firestore_client.batch()
        for _ in range(N_DOCS):
            batch.set(
                coll_ref.document(),
                {
                    "field": [] if coll_name == "performance_test_empty_array" else ["aa"],
                }
            )
        batch.commit()
        end = time.time()
        elapsed = round(end-start, 2)
        print(f"{elapsed} seconds to create {N_DOCS} documents")

        # Read data
        print("Fetching data")
        start = time.time()
        docs = coll_ref.get()
        end = time.time()
        elapsed = round(end-start, 2)
        print(f"{elapsed} seconds to fetch {N_DOCS} documents")
```
Leading to the following result:
```
performance_test_empty_array
Creating data
0.67 seconds to create 500 documents
Fetching data
10.9 seconds to fetch 500 documents

performance_test_not_empty_array
Creating data
0.65 seconds to create 500 documents
Fetching data
0.28 seconds to fetch 500 documents
```

### Issue analysis 
To demonstrate the issue with empty arrays, we created a collection containing 500 documents each having a single field as follows:

firestore_empty_arrays

We then fetched the data with the following code:
```python
from google.cloud import firestore
import time
import os

N_DOCS = 500

if __name__ == '__main__':

    os.environ['GOOGLE_APPLICATION_CREDENTIALS'] = "./gcp_sa_creds-dev.json"
    firestore_client = firestore.Client()

    coll_ref = firestore_client.collection("performance_test_empty_array")

    # Read data
    print("Fetching data")
    start = time.time()
    docs = coll_ref.get()
    end = time.time()
    elapsed = round(end-start, 2)
    print(f"{elapsed} seconds to fetch {N_DOCS} documents")
```
This required 9.76 seconds. By profiling the python code we understood that this issue was related to an extremely high
number of calls (15,295,500 times) to the `deepcopy` function of the Python standard library, as it can be seen
in the following profiling output:

profiler_500docs_empty_array

call_graph_empty_arrays

The call graph showed us that a lot of calls to the `deepcopy` function were caused by the `_pb_type` function defined in the  **RepeatedComposite** class (contained in "proto.marshal.collections.repeated") used to infer the protocol buffer type for a sequence. The reason for this is the following snippet taken from the "_pb_type" function:

```proto.marshal.collections.repeated:RepeatedComposite._pb_type```
```python
[...]
# We have no members in the list.
# In order to get the type, we create a throw-away copy and add a
# blank member to it.
canary = copy.deepcopy(self.pb).add()
return type(canary)
```
which performs a `deepcopy` of the `self.pb` element just to add an element in order to infer the type of the sequence.

### Proposed solution
By replacing the above snippet with the following snippet we can bypass the `deepcopy`:
```python
[...]
# We have no members in the list.
return self.pb._message_descriptor._concrete_class
```
That is, we get the type of the protocol buffer from the protected attributes. With this change, fetching the 500 documents took 0.24 seconds compared to 9.76 seconds with the original snippet. In fact, if we profile the code with the newly introduced change, we see that the number of calls to the `deepcopy` function went from 15,295,500 to 1,500.

### Tests
All tests pass after the proposed solution was implemented.
tests_passing

### Next steps
As soon as the maintainers have validated (or improved upon) our proposed solution, we can readily submit a PR to address this issue.

@zoncrd @lucastanzani

GabrieleMazzola:
Closes #224.

### DIFF
--- a/proto/marshal/collections/repeated.py
+++ b/proto/marshal/collections/repeated.py
@@ -104,11 +104,8 @@ class RepeatedComposite(Repeated):
         if len(self.pb) > 0:
             return type(self.pb[0])
 
-        # We have no members in the list.
-        # In order to get the type, we create a throw-away copy and add a
-        # blank member to it.
-        canary = copy.deepcopy(self.pb).add()
-        return type(canary)
+        # We have no members in the list, so we get the type from the attributes.
+        return self.pb._message_descriptor._concrete_class
 
     def __eq__(self, other):
         if super().__eq__(other):
--- a/proto/marshal/collections/repeated.py
+++ b/proto/marshal/collections/repeated.py
@@ -105,7 +105,16 @@ class RepeatedComposite(Repeated):
             return type(self.pb[0])
 
         # We have no members in the list, so we get the type from the attributes.
-        return self.pb._message_descriptor._concrete_class
+        if hasattr(self.pb, "_message_descriptor") and hasattr(
+            self.pb._message_descriptor, "_concrete_class"
+        ):
+            return self.pb._message_descriptor._concrete_class
+
+        # Fallback logic in case attributes are not available
+        # In order to get the type, we create a throw-away copy and add a
+        # blank member to it.
+        canary = copy.deepcopy(self.pb).add()
+        return type(canary)
 
     def __eq__(self, other):
         if super().__eq__(other):


GabrieleMazzola:
@googlebot I signed it!

software-dov:
It looks like the cpp runtime is causing problems because the type it exposes doesn't have the same python-visible fields. There's a couple of ways we can try solving this problem:
1) Complain to the protobuf folks and get a generic, runtime independent mechanism for determining the type of the elements of a repeated field.
2) Have a fast path (proposed change) for the python runtime and try/except to the deepcopy and canary.
3) Continue to dig deep for a field/mechanism that's faster than deep copy and works on the cpp runtime types.
4) Something like this (also passes the tests, but I'm not entirely sure it's 100% safe):
```diff
--- a/proto/marshal/marshal.py
+++ b/proto/marshal/marshal.py
@@ -155,12 +155,12 @@ class BaseMarshal:
         # Return a view around it that implements MutableSequence.
         value_type = type(value)  # Minor performance boost over isinstance
         if value_type in compat.repeated_composite_types:
-            return RepeatedComposite(value, marshal=self)
+            return RepeatedComposite(value, marshal=self, proto_type=proto_type)
         if value_type in compat.repeated_scalar_types:
             if isinstance(proto_type, type):
                 return RepeatedComposite(value, marshal=self, proto_type=proto_type)
             else:
-                return Repeated(value, marshal=self)
+                return Repeated(value, marshal=self, proto_type=proto_type)
 ```

**Comment by busunkim96** [comment]
How about we go with 2 to unblock folks, and follow up with the protobuf folks to figure out which of 1, 3, or 4 is the best option? Is there a person we can tag in to this PR?

I am somewhat uncomfortable with relying on a private attribute from `protobuf`, but it seems OK for the short term with a fallback (option 2).

**Comment by crwilcox** [comment]
Also +1 to option 2.

**Comment by craiglabenz** [comment]
Do we have a sense of what percentage of libraries would benefit from this change, versus trip the try/catch and actually run _slower_? I certainly don't, but I do know that `python-datastore` raises an AttributeError and would be further slowed.

As I'm thinking about this further, I suppose the answer is to only bump `proto-plus` versions in libraries that have been confirmed to work smoothly with this change?

I also don't know anything about the cpp layer and would love to learn more.

**Comment by busunkim96** [comment]
@craiglabenz The cpp option is opt-in according to teh documentation, although I saw some comments on issues that seemed to suggest it was default.

I think it would be helpful to clarify with a protobuf person who knows for sure.

**Comment by crwilcox** [comment]
We could also check for the attr without a try/catch if we are concerned about the cost of that.

**Comment by craiglabenz** [comment]
I am increasingly suspicious that this deepcopy is not the guilty party. In testing the implications of this change and the snippet proposed in [the comment above](https://github.com/googleapis/proto-plus-python/pull/226#issuecomment-819071172), I just discovered that all of python-datastore's slowness must be coming from elsewhere, because my scratch code that recreates said slowness never reaches that deepcopy.

**Comment by software-dov** [comment]
Responding to a bunch of things all at once.

> Do we have a sense of what percentage of libraries would benefit from this change, versus trip the try/catch and actually run slower?

There's a couple misconceptions in this question. The python/cpp protobuf runtimes are agnostic of the client libraries themselves: they are determined by platform and environment variables. The cpp runtime is much faster in certain real world workloads, but I don't have a good idea of what the breakdown in user environments between the cpp and py runtimes is.
Also, in most cases, `try`/`catch` is _fast_; especially if the straightforward path is more common, it is often faster to handle the common case in `try` and handle the uncommon in `except` than to do normal `if` `else` chains.

> I saw some comments on issues that seemed to suggest [cpp] was default.

This is a little bit complicated. If a cpp runtime for a particular platform exists and is installed, it is now the default version. We can see this via code archaeology in [api_implementation.py](https://github.com/protocolbuffers/protobuf/blob/master/python/google/protobuf/internal/api_implementation.py):40.

> I am increasingly suspicious that this deepcopy is not the guilty party.

That's entirely possible. Designing a good, helpful benchmark that mimics real world uses is tricky. I gave up after about two hours trying to recreate the benchmark linked in the issue for this PR. On the other hand, `RepeatedField` serialization is a known real-world hotspot and have not been highly optimized. Trying to resolve this issue seems like a good idea, especially if it can be done without to much effort or stress.

**Comment by software-dov** [comment]
What are people's required timelines for dealing with this performance regression? Clearly sooner is better, but is anything being blocked, or has conversion to the new client libraries been halted, or anything?

File: proto/marshal/collections/repeated.py [hunk 1]
```diff
@@ -104,11 +104,8 @@ def _pb_type(self):
         if len(self.pb) > 0:
             return type(self.pb[0])
 
-        # We have no members in the list.
-        # In order to get the type, we create a throw-away copy and add a
-        # blank member to it.
-        canary = copy.deepcopy(self.pb).add()
-        return type(canary)
+        # We have no members in the list, so we get the type from the attributes.
+        return self.pb._message_descriptor._concrete_class
```
craiglabenz, line 108: I ported this change to googleapis/python-datastore and discovered that these attributes are not universally available.

We might need to do something like this, but then assuming an equivalent change, this would LGTM.

```py
if hasattr(self.pb, '_message_descriptor') and hasattr(self.pb._message_descriptor, '_concrete_class'):
    return self.pb._message_descriptor._concrete_class

canary = copy.deepcopy(self.pb).add()
return type(canary)
```

craiglabenz (changes requested):

proto/marshal/collections/repeated.py
software-dov: > these attributes are not universally available.

This is the difference between the cpp and python protobuf runtimes. The concrete type of `self.pb` is different depending on which runtime is being used; it is **not** dependent on the API or the client library itself.

The `_message_descriptor` attribute is apparently considered an implementation detail of the python based protobuf runtime.

craiglabenz: I partially followed that (due to my ongoing ignorance of the cpp layer), but to clarify, are you suggesting that this line of code will work everywhere?

crwilcox:
@software-dov I know of at least one instance where a user has reverted to the previous major version due to this regression relative to the monolith generator.

proto/marshal/collections/repeated.py
software-dov: The change will work for all client libraries **IFF** the application process is using the python protobuf runtime.

This is the relevant chunk of the tech stack:
* User application
* Client library manual layer (optional. Firestore has a manual layer, Texttospeech does not)
* Generated client library, aka GAPIC
* Proto plus runtime, i.e. this repo
* General protobuf runtime, either written in pure python or in cpp as a python extension, which is determined at runtime. Proto plus _should_ be agnostic about which which is being used.

The lowest layer in the above is library is preventing the general fix from merging. The two different implementations provide different unofficial APIs.

crwilcox: @software-dov can you expand on the pure python vs cpp protobuf runtime? How would I get to each of these? Which one is used by cloud libraries?

software-dov: > Which one is used by cloud libraries?

It is chosen dynamically at runtime based on the platform (linux, macos X amd64, aarch64), version of protobuf installed, and environment variables. To be strictly general and not break cloud, any solution must be compatible with both. I would imagine, based on the environment they're running in, that most user applications tend to use the cpp runtime.

The protobuf runtime is responsible for memory layout, serialization and deserialization, and message introspection. It is the code that allocates memory and performs host-to-network and network-to-host bit conversion.

busunkim96:
I opened https://groups.google.com/g/protobuf/c/pYcq-UBixqU to get for some clarification from the Protobuf folks on the specific situations protobuf defaults to the C++ implementation (as well as when the change was made).

If you're curious, you can figure out which implementation protobuf is using with this snippet. I get 'cpp' on a fresh install on my linux workstation.

```py
from google.protobuf.internal import api_implementation
print(api_implementation.Type())
```

Note that protobuf discourages using `api_implementation.Type()` ([comment](https://github.com/protocolbuffers/protobuf/blob/6051341ab21b4e1f75e91710ab1a8cc8fadda521/python/google/protobuf/internal/api_implementation.py#L138-L143)).  That leads me to think think they'd be open to adding a way to fetch the type of elements in a repeated field agnostic of the implementation type. I'll open another issue to start that discussion.


Are folks alright with the code in this PR coupled with a fallback to deepcopy to handle the CPP case (`try`/`except`, `if`/`else`, attribute checking)? It sounds like it made a significant difference for the PR author in a real world scenario.  I imagine it will be a net positive for some folks and wouldn't make things worse for anyone else.

stefanbluegrid:
Hello guys, what is happening with this PR? If it can be fixed with simple `try/except` or `if` statement, I don't see reason for not being merged.

parthea:
Waiting for https://github.com/googleapis/proto-plus-python/pull/312

parthea (approved):

PR #230, opened 2021-05-04
[WIP] Protobuf Performance Refactor

### LINKED ISSUE #222: proto-plus introduces substantial overhead to interactions with pb.
Profiled/Verified: https://github.com/googleapis/python-pubsub/pull/158

Suspected: https://github.com/googleapis/python-datastore/issues/145

It seems that the use of proto-plus is introducing somewhere between 40-150% overhead to client libraries when using proto plus.

craiglabenz:
Aims to speed up naive usage on proto-plus-python as a wrapper around primitive protobuf objects.

This PR replaces attribute misses on proto-plus objects, which wound up in the `__getattr__` override to query the hidden protobuf object directly, with descriptors which do the same inside a caching layer to optimize repeated access. This descriptor/caching layer also aims to avoid unnecessary marshaling altogether when circumstances allow, further reducing the runtime of basic tasks.

---

The following is a performance chart of several simple tasks executed against the same proto definitions, first as raw protobuf objects and again as a proto-plus object. The two definitions are:

_Note: All tasks were run 10,000 times._

| Generate a(n)... | pb2 | proto-plus (before refactor) | proto-plus (with refactor) | refactor impact |
|---|---|---|---|---|
| empty instance | 3ms | 10ms | 7ms | ⬇️ 30% 🟢 |
| populated instance | 9ms | 168ms | 118ms | ⬇️ 30% 🟢 |
| populated nested objs | 38ms | 211ms | 163ms | ⬇️ 23% 🟢 |
| populated nested objs (1x attribute reads) | 39ms | 326ms | 419ms | ⬆️ 29% 🔴 |
| populated nested objs (10x attribute reads) | 101ms | 2009ms | 706ms | ⬇️ 65% 🟢 |
| populated w serialization | 20ms | 195ms | 280ms | ⬆️ 43% 🔴 |
| populated w lots of fields | 25ms | 624ms | 440ms | ⬇️ 30% 🟢 |
| MapFields populated during init | 124ms | 799ms | 626ms | ⬇️ 21% 🟢 |
| MapFields populated after init | -- | 1079ms | 369ms | ⬇️ 66% 🟢 |
| deserializing objects | 116ms | 204ms | 180ms | ⬇️ 12% 🟢 |

---

When used as a dependency in python-datastore, the following proto-plus versions have these performance hydrating loaded 500 entities (network I/O is excluded):

| datastore version | notes | before googleapis/proto-plus-python#230 | with googleapis/proto-plus-python#230 |
|--|--|--|--|
| 1.15.3 | before proto-plus | 360ms | _N/A_ |
| 2.0.0 | with proto-plus | 675ms | 800ms |
| 2.1.2 | proto-plus usage optimized <br> via googleapis/python-datastore#155) | 550ms | 750ms |

---

Resolves #222

### DIFF
--- a/proto/fields.py
+++ b/proto/fields.py
@@ -136,10 +136,6 @@ class Field:
         return self.mcls_data["package"]
 
     @property
-    def inner_pb_type(self):
-        return None
-
-    @property
     def pb_type(self):
         """Return the composite type of the field, or None for primitives."""
         # For enums, return the Python enum.
@@ -263,10 +259,6 @@ class MapField(Field):
     def __init__(self, key_type, value_type, *, number: int, message=None, enum=None):
         super().__init__(value_type, number=number, message=message, enum=enum)
         self.map_key_type = key_type
-
-    @property
-    def inner_pb_type(self):
-        return
 
 
 class _FieldDescriptor:
--- a/proto/fields.py
+++ b/proto/fields.py
@@ -157,11 +157,11 @@ class Field:
             return False
         return True
 
-    def can_set_natively(self, val: Any) -> bool:
+    @property
+    def can_set_natively(self) -> bool:
         if self.proto_type == ProtoType.MESSAGE and self.message == struct_pb2.Value:
             return False
         return True
-        # return self.pb_type is None and not self.repeated
 
     def contribute_to_class(self, cls, name: str):
         """Attaches a descriptor to the top-level proto.Message class, so that attribute
@@ -176,13 +176,13 @@ class Field:
             # Bytes are accepted for string values, but strings are not accepted for byte values.
             # This is an artifact of older Python2 implementations.
             set_coercion = self._bytes_to_str
-        if self.pb_type == timestamp_pb2.Timestamp:
+        elif self.pb_type == timestamp_pb2.Timestamp:
             set_coercion = self._timestamp_to_datetime
-        if self.proto_type == ProtoType.MESSAGE and self.message == duration_pb2.Duration:
+        elif self.proto_type == ProtoType.MESSAGE and self.message == duration_pb2.Duration:
             set_coercion = self._duration_to_timedelta
-        if self.proto_type == ProtoType.MESSAGE and self.message == wrappers_pb2.BoolValue:
+        elif self.proto_type == ProtoType.MESSAGE and self.message == wrappers_pb2.BoolValue:
             set_coercion = self._bool_value_to_bool
-        if self.enum:
+        elif self.enum:
             set_coercion = self._literal_to_enum
         setattr(cls, name, _FieldDescriptor(name, cls=cls, set_coercion=set_coercion))
 
@@ -281,7 +281,7 @@ class _FieldDescriptor:
         self.instance_attr_name = f'_cached_fields__{name}'
 
         # simple types coercion for setting attributes (for example, bytes -> str)
-        self._set_coercion = set_coercion or _FieldDescriptor._noop
+        self._set_coercion: Optional[Callable] = set_coercion
         self.cls = cls
 
     @property
@@ -318,8 +318,8 @@ class _FieldDescriptor:
             delattr(instance, field_name)
 
     def __set__(self, instance, value):
-        """Called whenever a value is assigned to a proto.Field attribute on an instantiated
-        proto.Message object.
+        """Called whenever a value is assigned to a `proto.Field` attribute on an instantiated
+        `proto.Message` object.
 
         Usage:
 
@@ -329,10 +329,10 @@ class _FieldDescriptor:
             my_message = MyMessage()
             my_message.name = "Frodo"
 
-        In the above scenario, `__set__` is called with "Frodo" passed as `value` and `my_instance`
-        passed as `instance`.
+        In the above scenario, `__set__` is called with "Frodo" passed as `value` and
+        `my_message` passed as `instance`.
         """
-        value = self._set_coercion(value)
+        value = self._set_coercion(value) if self._set_coercion is not None else value
         value = self._hydrate_dicts(value)
 
         # Warning: `always_commit` is hacky!
@@ -342,7 +342,7 @@ class _FieldDescriptor:
         # instances receive attribute updates, immediately syncing those values to the underlying
         # pb2 instance is sufficient.
         always_commit: bool = getattr(instance, '_always_commit', False)
-        if always_commit or not self.field.can_set_natively(value):
+        if always_commit or not self.field.can_set_natively:
             pb_value = instance._meta.marshal.to_proto(self.field.pb_type, value)
             _pb = instance._meta.pb(**{self.original_name: pb_value})
             instance._pb.ClearField(self.original_name)
@@ -382,13 +382,13 @@ class _FieldDescriptor:
         value = getattr(instance, self.instance_attr_name, None)
         if self.field.can_get_natively and value is not None:
             return value
-        else:
-            # For the most part, only primitive values can be returned natively, meaning
-            # this is either a Message itself, in which case, since we're dealing with the
-            # underlying pb object, we need to sync all deferred fields.
-            # This is functionally a no-op if no fields have been deferred.
-            if hasattr(value, '_update_pb'):
-                value._update_pb()
+
+        # For the most part, only primitive values can be returned natively, meaning
+        # this is either a Message itself, in which case, since we're dealing with the
+        # underlying pb object, we need to sync all deferred fields.
+        # This is functionally a no-op if no fields have been deferred.
+        if hasattr(value, '_update_pb'):
+            value._update_pb()
 
         pb_value = getattr(instance._pb, self.original_name, None)
         value = instance._meta.marshal.to_python(self.field.pb_type, pb_value, absent=self.original_name not in instance)
@@ -402,10 +402,6 @@ class _FieldDescriptor:
         instance._pb.ClearField(self.original_name)
         if self.original_name in getattr(instance, '_stale_fields', []):
             instance._stale_fields.remove(self.original_name)
-
-    @staticmethod
-    def _noop(val):
-        return val
 
 
 __all__ = (
--- a/proto/fields.py
+++ b/proto/fields.py
@@ -15,9 +15,10 @@
 import datetime
 from enum import EnumMeta
 from re import L
-from proto.marshal.rules import wrappers
 from proto.datetime_helpers import DatetimeWithNanoseconds
+from proto.marshal.rules import wrappers
 from proto.marshal.rules.dates import DurationRule
+from proto.utils import cached_property
 from typing import Any, Callable, Optional, Union
 
 from google.protobuf import descriptor_pb2
@@ -186,31 +187,23 @@ class Field:
             set_coercion = self._literal_to_enum
         setattr(cls, name, _FieldDescriptor(name, cls=cls, set_coercion=set_coercion))
 
-    @property
+    @cached_property
     def reverse_enum_map(self):
         """Helper that allows for constant-time lookup on self.enum, used to hydrate
         primitives that are supplied but which stand for their official enum types.
 
         This is used when a developer supplies the literal value for an enum type (often an int).
         """
-        if not self.enum:
-            return None
-        if not getattr(self, '_reverse_enum_map', None):
-            self._reverse_enum_map = {e.value: e for e in self.enum}
-        return self._reverse_enum_map
+        return {e.value: e for e in self.enum} if self.enum else None
 
-    @property
+    @cached_property
     def reverse_enum_names_map(self):
         """Helper that allows for constant-time lookup on self.enum, used to hydrate
         primitives that are supplied but which stand for their official enum types.
 
         This is used when a developer supplies the string value for an enum type's name.
         """
-        if not self.enum:
-            return None
-        if not getattr(self, '_reverse_enum_names_map', None):
-            self._reverse_enum_names_map = {e.name: e for e in self.enum}
-        return self._reverse_enum_names_map
+        return {e.name: e for e in self.enum} if self.enum else None
 
     def _literal_to_enum(self, val: Any):
         if isinstance(val, self.enum):
@@ -274,13 +267,27 @@ class _FieldDescriptor:
     long-lived, and thus information about which fields are stale would be lost if syncing
     was left for serialization time.
     """
+
+    # Namespace for attributes where we will store the Pythonic values of
+    # various `proto.Field` classes on instantiated `proto.Message` objects.
+    # For example, in the following scenario, this attribute is involved in
+    # saving the value "Homer Simpson" to `my_message._cached_fields__name`.
+    #
+    #   class MyMessage(proto.Message):
+    #       name = proto.Field(proto.STRING, ...)
+    #
+    #    my_message = MyMessage()
+    #    my_message.name = "Homer Simpson"  # saves to `_cached_fields__name`
+    cached_fields_prefix = '_cached_fields__'
+
     def __init__(self, name: str, *, cls, set_coercion: Optional[Callable] = None):
         # something like "id". required whenever reach back to the pb2 object.
         self.original_name = name
         # something like "_cached_id"
-        self.instance_attr_name = f'_cached_fields__{name}'
+        self.instance_attr_name = f'{self.cached_fields_prefix}{name}'
 
-        # simple types coercion for setting attributes (for example, bytes -> str)
+        # simple types coercion for setting attributes
+        # (e.g., bytes -> str if our type is string, but we are supplied bytes)
         self._set_coercion: Optional[Callable] = set_coercion
         self.cls = cls
 
@@ -379,8 +386,8 @@ class _FieldDescriptor:
         if instance is None:
             return self.original_name
 
-        value = getattr(instance, self.instance_attr_name, None)
-        if self.field.can_get_natively and value is not None:
+        value = getattr(instance, self.instance_attr_name, _none)
+        if self.field.can_get_natively and value is not _none:
             return value
 
         # For the most part, only primitive values can be returned natively, meaning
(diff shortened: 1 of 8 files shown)


craiglabenz:
Note for code reviewers: There are a few spots in this PR that are pseudo-experiments, and they're decorated with GitHub comments, but please feel free to critique it all!

proto/marshal/marshal.py
```diff
@@ -122,6 +127,80 @@ def register_rule_class(rule_class: type):
...
 
+    @staticmethod
+    def _throw_type_error(primitive):
+        raise TypeError(f"Unacceptable value of type {type(primitive)}")
+
+    @staticmethod
+    def _throw_value_error(proto_type, primitive):
+        raise ValueError(f"Unacceptable value {type(primitive)} for type {proto_type}")
+
+    def validate_primitives(self, field, primitive):
```
craiglabenz, line 138: This function is unfortunate, but is the only way to recreate immediate validation when attempting something like `my_obj.some_string_field = 123`.

proto/message.py
```diff
@@ -269,6 +287,14 @@ def __new__(mcls, name, bases, attrs):
     def __prepare__(mcls, name, bases, **kwargs):
         return collections.OrderedDict()
 
+    def add_to_class(cls, name, value):
```
craiglabenz, line 290: Note: This whole `contribute_to_class` pattern is heavily (or, more like entirely) [borrowed from Django](https://github.com/django/django/blob/a77c9a4229cfef790ec18001b2cd18bd9c4aedbc/django/db/models/base.py#L324-L328).

craiglabenz (commented):
Some (hopefully) helpful mile-markers for code review.

proto/fields.py
```diff
@@ -155,6 +261,153 @@ def __init__(self, key_type, value_type, *, number: int, message=None, enum=None
...
+        In the above scenario, `__get__` is called with "my_message" passed as `instance`.
+        """
+        # If `instance` is None, then we are accessing this field directly
+        # off the class itself instead of off an instance.
+        if instance is None:
+            return self.original_name
+
+        value = getattr(instance, self.instance_attr_name, None)
+        if self.field.can_get_natively and value is not None:
+            return value
```
tseaver, line 384: I don't have enough context here to know whether `instance` can ever have an actual attr for the field with value `None`:  my first inclination was to suggest a marker such as `_nonesuch`:
```suggestion
        value = getattr(instance, self.instance_attr_name, _nonesuch)
        if self.field.can_get_natively and value is not _nonesuch:
            return value
```

tseaver (commented):
Ran out of time today to look at more than `proto/fields.py`.  Hope the review is helpful anyway.

busunkim96:
@craiglabenz Thank you for putting this together!

@software-dov Could you take a look?

proto/message.py
```diff
@@ -31,6 +33,13 @@
 from proto.primitives import ProtoType
 
 
+def _has_contribute_to_class(value):
+    # Only call contribute_to_class() if it's bound.
+    return not inspect.isclass(value) and hasattr(value, 'contribute_to_class')
+
+class MapValueMessage:
+    isMapValue: bool = True
```
software-dov, line 41: This seems to be unused. Is it part of WIP, or is it cruft?

proto/message.py
```diff
@@ -446,6 +478,14 @@ class Message(metaclass=MessageMeta):
     """
 
     def __init__(self, mapping=None, *, ignore_unknown_fields=False, **kwargs):
+
+        self._cached_pb = None
+        # Tracks any new values have had their serialization deferred, and thus whether we
+        # are ready to serialize immediately, or need to sync from the instance back to the
+        # underlying `_pb`
+        self._stale_fields: List[str] = []
```
software-dov, line 486: I think `_stale_fields` should be a set instead of a list. It's setting a de-facto sticky dirty bit on a field and then providing iteration.

software-dov (commented):
First pass, going to need a few others to fully understand what's going on.

software-dov:
Can you provide a link to your benchmark code? I've had a hard time creating good benchmarks that mimic real world message manipulation patterns.

proto/fields.py
```diff
@@ -140,6 +151,101 @@ def pb_type(self):
...
+    def can_get_natively(self) -> bool:
+        if self.proto_type == ProtoType.MESSAGE and self.message == struct_pb2.Value:
+            return False
+        return True
+
+    def can_set_natively(self, val: Any) -> bool:
+        if self.proto_type == ProtoType.MESSAGE and self.message == struct_pb2.Value:
+            return False
+        return True
+        # return self.pb_type is None and not self.repeated
```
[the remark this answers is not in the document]
craiglabenz: Ha, whoops. Deleted.

proto/fields.py
```diff
@@ -140,6 +151,101 @@ def pb_type(self):
...
+        reads and writes can be specially handled in `_FieldDescriptor.__get__` and
+        `FieldDescriptor.__set__`.
+
+        Also contains hooks for write-time type-coersion to translate special cases between
+        pure Pythonic objects and pb2-compatible structs or values.
+        """
+        set_coercion = None
+        if self.proto_type == ProtoType.STRING:
+            # Bytes are accepted for string values, but strings are not accepted for byte values.
+            # This is an artifact of older Python2 implementations.
```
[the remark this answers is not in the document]
craiglabenz: My worry is that code out there is still (maybe pointlessly?) making use of the fact that bytes are accepted here. I don't think we could easily get rid of this without a deprecation warning phase.

That said, I'm happy to rethink it if I've overlooked a way to rule out this scenario.

proto/fields.py
```diff
@@ -140,6 +151,101 @@ def pb_type(self):
...
+            set_coercion = self._bytes_to_str
+        if self.pb_type == timestamp_pb2.Timestamp:
```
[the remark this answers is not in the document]
craiglabenz: Good call - changed!

proto/fields.py
```diff
@@ -155,6 +261,153 @@ def __init__(self, key_type, value_type, *, number: int, message=None, enum=None
...
+            class MyMessage(proto.Message):
+                name = proto.Field(proto.STRING, number=1)
+
+            my_message = MyMessage()
+            my_message.name = "Frodo"
+
+        In the above scenario, `__set__` is called with "Frodo" passed as `value` and `my_instance`
+        passed as `instance`.
+        """
+        value = self._set_coercion(value)
```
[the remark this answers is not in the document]
craiglabenz: I like this idea! Adding.

proto/fields.py
```diff
@@ -140,6 +151,101 @@ def pb_type(self):
             return self.message.pb()
         return self.message
 
+    @property
+    def can_get_natively(self) -> bool:
+        if self.proto_type == ProtoType.MESSAGE and self.message == struct_pb2.Value:
+            return False
+        return True
+
+    def can_set_natively(self, val: Any) -> bool:
```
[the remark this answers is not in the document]
craiglabenz: I think it was briefly needed, then refactored out. Changed to a `@property`.

proto/fields.py
```diff
@@ -155,6 +261,153 @@ def __init__(self, key_type, value_type, *, number: int, message=None, enum=None
...
+        Usage:
+
+            class MyMessage(proto.Message):
+                name = proto.Field(proto.STRING, number=1)
+
+            my_message = MyMessage()
+            my_message.name = "Frodo"
+
+        In the above scenario, `__set__` is called with "Frodo" passed as `value` and `my_instance`
+        passed as `instance`.
```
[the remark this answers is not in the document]
craiglabenz: Great catch!

proto/marshal/collections/maps.py
```diff
@@ -28,7 +29,7 @@ class MapComposite(collections.abc.MutableMapping):
     def _pb_type(self):
         """Return the protocol buffer type for this sequence."""
         # Huzzah, another hack. Still less bad than RepeatedComposite.
-        return type(self.pb.GetEntryClass()().value)
+        return type(self.self.pb.GetEntryClass()().value)
```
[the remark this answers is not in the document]
craiglabenz: Sure was 😬

proto/message.py
craiglabenz: That's a good point. Will change.

proto/fields.py
```diff
@@ -140,6 +151,101 @@ def pb_type(self):
             return self.message.pb()
         return self.message
 
+    @property
+    def can_get_natively(self) -> bool:
+        if self.proto_type == ProtoType.MESSAGE and self.message == struct_pb2.Value:
```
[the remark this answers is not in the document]
craiglabenz: Nothing I've seen so far suggests this will be a hotspot, so I'm going to resolve this conversation.

proto/fields.py
```diff
@@ -155,6 +261,153 @@ def __init__(self, key_type, value_type, *, number: int, message=None, enum=None
...
+    their various implementations within pb2 (which for our purposes is mostly a black box)
+    sometimes mandate immediate syncing. This is usually because proto-plus objects are not
+    long-lived, and thus information about which fields are stale would be lost if syncing
+    was left for serialization time.
+    """
+    def __init__(self, name: str, *, cls, set_coercion: Optional[Callable] = None):
+        # something like "id". required whenever reach back to the pb2 object.
+        self.original_name = name
+        # something like "_cached_id"
+        self.instance_attr_name = f'_cached_fields__{name}'
```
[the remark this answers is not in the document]
craiglabenz: Added explanatory comments.

proto/marshal/marshal.py
```diff
@@ -122,6 +127,80 @@ def register_rule_class(rule_class: type):
...
+        on instantiated proto.Message objects. This is required to recreate the immediacy
+        of ValueError and TypeError checks that the pb2 layer made, but which are now
+        deferred because syncing to the pb2 layer itself is often entirely deferred.
+        """
+        proto_type = field.proto_type
+        if field.repeated:
+            if primitive and not isinstance(primitive, (list, Repeated, RepeatedComposite,)):
+                self._throw_type_error(primitive)
+            if primitive:
+                primitive = primitive[0]
```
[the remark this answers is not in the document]
craiglabenz: Do any fancier list types (like `RepeatedCompositeField`, or its cousins) fail to resolve to `False` when empty? I'll add a length check and comment to be safe.

proto/marshal/marshal.py
```diff
@@ -122,6 +127,80 @@ def register_rule_class(rule_class: type):
...
+                self._throw_value_error(proto_type, primitive)
+        elif proto_type == proto.SFIXED64 and _type != int:
+            self._throw_type_error(primitive)
+        elif proto_type == proto.SINT32:
+            if _type != int:
+                self._throw_type_error(primitive)
+            if primitive < min_int_32 or primitive > max_int_32:
+                self._throw_value_error(proto_type, primitive)
+        elif proto_type == proto.SINT64 and _type != int:
+            self._throw_type_error(primitive)
```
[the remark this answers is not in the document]
craiglabenz: Interesting idea. To make sure I didn't miss anything, I'd followed the order of prototype definitions, but in light of this thought, I've now moved `string` and `bool` to the top of the list alongside the other more common primitive types.

proto/message.py
craiglabenz: Whoops, thought I deleted that.

proto/marshal/marshal.py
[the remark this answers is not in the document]
software-dov: I believe that _now_ empty repeated composites are guaranteed to be falsy. This used to not be the case.
Edit but not really: A quick experiment confirms that `RepeatedCompositeContainer`s are in fact falsy.

proto/message.py
```diff
@@ -572,65 +606,64 @@ def __eq__(self, other):
...
-        marshal = self._meta.marshal
-        pb_type = self._meta.fields[key].pb_type
-        pb_value = marshal.to_proto(pb_type, value)
+        for field_name, field in self._meta.fields.items():
+            if field.proto_type == proto.MESSAGE:
+                obj = getattr(self, field_name, None)
+                if obj and hasattr(obj, '_update_pb'):
+                    obj._update_pb()
+                if obj and hasattr(obj, '_update_nested_pb'):
+                    obj._update_nested_pb()
```
software-dov, line 649: It took me a few readthroughs to understand why a field value may be a message and not have `_update_pb` or `_update_nested_pb`. I bet that if we changed the conditional logic a bit it might be more clear.
```python
if obj and isinstance(obj, Message):
    obj._update_pb()
    obj._update_nested_pb()
```
I still need to mull over this logic a little bit. It seems like we might be able to skip fields under certain circumstances. Also, I'm not sure if the ordering matters. I'll need to do some experiments and come up with examples and or counterexamples.

proto/message.py
```diff
@@ -539,6 +577,8 @@ def __contains__(self, key):
             bool: Whether the field's value corresponds to a non-empty
                 wire serialization.
         """
+        if key in getattr(self, '_stale_fields', set()):
+            return True
```
software-dov, line 581: Aren't we guaranteed by the constructor that `_stale_fields` will exist?

proto/message.py
```diff
@@ -572,65 +606,64 @@ def __eq__(self, other):
...
+    def _pb(self, value):
+        self._cached_pb = value
+
+    def _mark_pb_stale(self, field_name: str):
+        """We often set fields on the proto-plus object (this), and do not immediately
+        sync them down to the underlying pb2 object. This mechanism tracks which
+        fields are stale in this way.
+        """
+        if not hasattr(self, '_stale_fields'):
+            self._stale_fields = set()
```
software-dov, line 632: Same here: I don't think we need to check `hasattr(self, '_stale_fields')`

proto/message.py
```diff
@@ -572,65 +606,64 @@ def __eq__(self, other):
...
+        self._stale_fields.add(field_name)
 
-        For well-known protocol buffer types which are marshalled, either
-        the protocol buffer object or the Python equivalent is accepted.
+    def _mark_pb_synced(self):
+        self._stale_fields = set()
```
software-dov, line 636: Minor optimiation: `set` has an explicit `clear` method that's a little faster.

proto/message.py
```diff
@@ -572,65 +606,64 @@ def __eq__(self, other):
...
+                if obj and hasattr(obj, '_update_pb'):
+                    obj._update_pb()
+                if obj and hasattr(obj, '_update_nested_pb'):
+                    obj._update_nested_pb()
+
+    def _update_pb(self):
+        """Loops over any stale fields and syncs them to the underlying pb2 object.
+        """
+        merge_params = {}
+        for field_name in getattr(self, '_stale_fields', set()):
```
software-dov, line 655: Same here: we can just take the existence of `_stale_fields` as given.

software-dov (commented):
Second review round, looking great!

proto/fields.py
```diff
@@ -140,6 +152,93 @@ def pb_type(self):
...
+    def can_get_natively(self) -> bool:
+        if self.proto_type == ProtoType.MESSAGE and self.message == struct_pb2.Value:
+            return False
+        return True
+
+    @property
+    def can_set_natively(self) -> bool:
+        if self.proto_type == ProtoType.MESSAGE and self.message == struct_pb2.Value:
+            return False
+        return True
```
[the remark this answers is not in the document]
craiglabenz: Ha, yep, I'd just noticed that and made the same change.

proto/fields.py
```diff
@@ -155,6 +254,175 @@ def __init__(self, key_type, value_type, *, number: int, message=None, enum=None
...
+    # saving the value "Homer Simpson" to `my_message._cached_fields__name`.
+    #
+    #   class MyMessage(proto.Message):
+    #       name = proto.Field(proto.STRING, ...)
+    #
+    #    my_message = MyMessage()
+    #    my_message.name = "Homer Simpson"  # saves to `_cached_fields__name`
+    cached_fields_prefix = '_cached_fields__'
+
+    def __init__(self, name: str, *, cls, set_coercion: Optional[Callable] = None):
```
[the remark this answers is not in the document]
craiglabenz: There used to be a no-op default, but @tseaver opined that this pattern might be cleaner. Either way, expanding on the comments.

proto/marshal/collections/maps.py
```diff
@@ -54,7 +55,10 @@ def __getitem__(self, key):
         # buffers will create the key if it does not exist.
         if key not in self:
             raise KeyError(key)
-        return self._marshal.to_python(self._pb_type, self.pb[key])
+        obj = self._marshal.to_python(self._pb_type, self.pb[key])
+        if isinstance(obj, proto.Message):
+            obj._always_commit = True
```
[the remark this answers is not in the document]
craiglabenz: Oh yes, I didn't realize that I'd only left an appropriate comment on the other end of this wire.

proto/message.py
craiglabenz: I had thought this would be the case, too, but the pattern `MyMessage.wrap(...)` uses the metaclass directly and skips the class's actual constructor, so in many cases it is (surprisingly) missing.

proto/message.py
craiglabenz: Same as above: our usage of `MyMessage.wrap()` bypasses the constructor 😲

proto/message.py
tseaver: FWIW, constructing a set just to get the `in` operator to return false on a missing attribute is maybe an unnecessary slowdown:  the (interned) `()` would work just as well, or you could create an empty `frozenset` at module scope for it.

proto/message.py
tseaver: LIkewise here, constructing a set is expensive compared to using an already created (or interned) value.

craiglabenz:
Closing this PR because it optimizes use cases that are different from what we see in client libraries like python-datastore and python-firestore.

Specifically, caching attributes is only helpful if they are read > 1 time, yet python-datastore converts all proto-plus instances into dictionary-like objects; meaning each field is read exactly once before all future reads are performed against that dictionary.

If you find this and would like to know more, reach out.

PR #232, opened 2021-06-07
feat: pass 'including_default_value_fields' through to 'Message.to_dict' method

### LINKED ISSUE #231: bug: Cannot toggle including_default_value_fields in to_dict method.
https://github.com/googleapis/proto-plus-python/blob/e25dd7d65273a834f66d3be0445b88011103dc81/proto/message.py#L399

```including_default_value_fields```  is always set to ```True``` in ```proto.Message.to_dict()``` method .

Its value should set from kwargs of ```to_dict()```  method , same as ```to_json()``` method .

Use of ```to_dict()``` method to convert GoogleAdsRow (GoogleAds API v7) to dict results in unwanted additional columns in results.

### PR DESCRIPTION
**Pull request by vishwaja13**
Solution to the following issue #231, should work as boolean value and according to the value the GoogleAdsRow default columns should or should not return in results.

If `including_default_value_fields`  is `False` default columns should not return and if `True` default values should return now.

Closes #231.

### DIFF
--- a/proto/message.py
+++ b/proto/message.py
@@ -375,7 +375,8 @@ class MessageMeta(type):
         return instance
 
     def to_dict(
-        cls, instance, *, use_integers_for_enums=True, preserving_proto_field_name=True
+        cls, instance, *, use_integers_for_enums=True, preserving_proto_field_name=True,
+        including_default_value_fields=True,
     ) -> "Message":
         """Given a message instance, return its representation as a python dict.
 
@@ -396,7 +397,7 @@ class MessageMeta(type):
         """
         return MessageToDict(
             cls.pb(instance),
-            including_default_value_fields=True,
+            including_default_value_fields=including_default_value_fields,
             preserving_proto_field_name=preserving_proto_field_name,
             use_integers_for_enums=use_integers_for_enums,
         )
--- a/proto/message.py
+++ b/proto/message.py
@@ -375,7 +375,11 @@ class MessageMeta(type):
         return instance
 
     def to_dict(
-        cls, instance, *, use_integers_for_enums=True, preserving_proto_field_name=True,
+        cls, 
+        instance, 
+        *, 
+        use_integers_for_enums=True, 
+        preserving_proto_field_name=True,
         including_default_value_fields=True,
     ) -> "Message":
         """Given a message instance, return its representation as a python dict.
@@ -389,6 +393,7 @@ class MessageMeta(type):
             preserving_proto_field_name (Optional(bool)): An option that
                 determines whether field name representations preserve
                 proto case (snake_case) or use lowerCamelCase. Default is True.
+            including_default_value_fields (Optional(bool)): An option that determines whether the default field values should be included in the results. Default is True.
 
         Returns:
             dict: A representation of the protocol buffer using pythonic data structures.
--- a/proto/message.py
+++ b/proto/message.py
@@ -380,7 +380,7 @@ class MessageMeta(type):
         *, 
         use_integers_for_enums=True, 
         preserving_proto_field_name=True,
-        including_default_value_fields=True,
+        including_default_value_fields=True
     ) -> "Message":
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by vishwaja13** [comment]
@googlebot I signed it!

**Comment by vishwaja13** [comment]
> @googlebot I fixed it.

**Comment by vishwaja13** [comment]
@googlebot I fixed it.

**Comment by busunkim96** [comment]
Hi! Thanks for the PR.

It looks like CLA bot expects your commits to be associated with an @gmail account but the commit has `Vishwaja@Vishwajas-MacBook-Air.local` instead. Could you edit your commit please? (see [stackoverflow post](https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-one-specific-commit) for help doing this).

File: proto/message.py [hunk 1]
```diff
@@ -375,7 +375,7 @@ def from_json(cls, payload, *, ignore_unknown_fields=False) -> "Message":
         return instance
 
     def to_dict(
-        cls, instance, *, use_integers_for_enums=True, preserving_proto_field_name=True
+        cls, instance, *, use_integers_for_enums=True, preserving_proto_field_name=True, including_default_value_fields=True,
```
tseaver, line 378: Please wrap:
```suggestion
        cls,
        instance,
        *,
        use_integers_for_enums=True,
        preserving_proto_field_name=True,
        including_default_value_fields=True,
```

Also, please add `including_default_value_fields` to the docstring below.

tseaver (changes requested):
Thanks very much for the patch!  In addition to the tweaks requested inline, this PR needs to add a test (in `tests/test_message.py`) to ensure that the `including_default_value_fields` value is passed correctly.

proto/message.py
```diff
@@ -375,7 +375,12 @@ def from_json(cls, payload, *, ignore_unknown_fields=False) -> "Message":
         return instance
 
     def to_dict(
-        cls, instance, *, use_integers_for_enums=True, preserving_proto_field_name=True
+        cls, 
+        instance, 
+        *, 
+        use_integers_for_enums=True, 
+        preserving_proto_field_name=True,
+        including_default_value_fields=True,
```
software-dov, line 383: The default for `MessageToDict` is `False`, and implicitly changing it here could constitute a breaking change. I'm okay with adding this option, but I think we should keep the default False.

software-dov (changes requested):

software-dov (approved):

proto/message.py
```diff
@@ -388,6 +393,7 @@ def to_dict(
             preserving_proto_field_name (Optional(bool)): An option that
                 determines whether field name representations preserve
                 proto case (snake_case) or use lowerCamelCase. Default is True.
+            including_default_value_fields (Optional(bool)): An option that determines whether the default field values should be included in the results. Default is True.
```
tseaver, line 396: Please wrap here as well:
```suggestion
            including_default_value_fields (Optional(bool)): An option that
            determines whether the default field values should be included in
            the results. Default is True.
```

tests/test_message.py
```diff
@@ -341,3 +346,5 @@ class Squid(proto.Message):
 
     with pytest.raises(TypeError):
         Mollusc.Squid.copy_from(m.squid, (("mass_kg", 20)))
+
+test_serialize_to_dict()
```
tseaver, line 350: Please remove the test invocation here.

tseaver (changes requested):

tests/test_message.py
tseaver, line 350:
```suggestion
```

vishwaja13:
> @googlebot I fixed it.

vishwaja13:
@googlebot I fixed it.

tseaver (approved):

vishwaja13:
cla process completed.

vishwaja13:
workflow style-check fixes in tests/test_message.py

PR #237, opened 2021-07-02
fix: add an option to define a "proto name" for a field

software-dov:
This is used for defining useful field masks, among other things.

### DIFF
--- a/proto/fields.py
+++ b/proto/fields.py
@@ -36,7 +36,8 @@ class Field:
         enum=None,
         oneof: str = None,
         json_name: str = None,
-        optional: bool = False
+        optional: bool = False,
+        proto_name: str = None,
     ):
         # This class is not intended to stand entirely alone;
         # data is augmented by the metaclass for Message.
@@ -63,6 +64,7 @@ class Field:
         self.json_name = json_name
         self.optional = optional
         self.oneof = oneof
+        self.proto_name = proto_name
 
         # Once the descriptor is accessed the first time, cache it.
         # This is important because in rare cases the message or enum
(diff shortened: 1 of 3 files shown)


BenRKarl:
[Here](https://github.com/googleapis/gapic-generator-python/pull/945#discussion_r663178372) you mentioned this would be a breaking change for Ads, what exactly would break?

software-dov:
The names of fields in the underlying proto would be changed by removing the trailing underscore. Anyone referencing fields in the underlying proto would get errors.
To clarify, the change in this PR plus the changes in the generated surface:
```python
class MyMessage(proto.Message):
    type_ = proto.Field(..., proto_name="type")

msg = MyMessage()
# Now an attribute error
MyMessage.pb(msg).type_ = "squid"
```

BenRKarl:
Ok I see, that makes sense, so for example:

```py
msg = MyMessage()

msg.type_ = "squid"

print(MyMessage.pb(msg).type)
>> squid
```

Though this is technically breaking, there are almost no scenarios where we recommend using the underlying (or as we have come to call them, "native") protos. Also, I think this feels correct semantically. Proto plus is an ergonomic layer with some minor compromises, the native protos are more structurally accurate but difficult to work with. 

Of the solutions discussed I would prefer this one. Curious what @busunkim96 thinks.

software-dov:
@busunkim96 What do you think of this solution instead of removing `type` from the list of reserved words?

busunkim96:
Thanks for putting this together! This makes sense to me and seems nicer than modifying the reserved words list.

I think it should be released in a new major version (with `fix!: ...`), since it is still technically breaking change to a public interface and it's impossible to know if folks are relying on the existing behavior.

I can't think of any ways this will break handwritten clients that wrap GAPICs, but it's probably worth getting input from other library maintainers. I put it on the agenda for next Tuesday's meeting.

BenRKarl:
From an Ads perspective - we'll regenerate the library and release it as a major version, which isn't a problem. Although it's technically breaking, it breaks an interface that we don't recommend using.

PR #248, opened 2021-09-24
fix: ensure enums are incomparable w other enum types

### LINKED ISSUE #247: 'IntEnum' values from different types compare equal
For instance:

```python
from google.cloud.firestore_v1.types import StructuredQuery

greater_than = StructuredQuery.FieldFilter.Operator.GREATER_THAN
is_null = StructuredQuery.UnaryFilter.Operator.IS_NULL

assert greater_than != is_null  # fails
```

This "false equivalence" is the source of https://github.com/googleapis/python-firestore/issues/429

tseaver:
Fixes #247.

### DIFF
--- a/proto/enums.py
+++ b/proto/enums.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import enum
+import functools
 
 from google.protobuf import descriptor_pb2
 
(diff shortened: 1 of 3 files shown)


proto/enums.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import enum
+import functools
```
software-dov, line 16: Nit: remove import, not used anywhere.

software-dov (approved):

proto/enums.py
tseaver: ba7a451

PR #255, opened 2021-09-29
fix: setting bytes field from python string base64 decodes before assignment

### LINKED ISSUE #249: Message With Bytes Field Won't Re-Serialize as dict
If a message has a bytes field, to_dict serializes the field as a base64 string, which cannot then be read back in to re-create the object.

#### Environment details

  - Programming language: python
  - OS: linux
  - Language runtime version: 3.9.7
  - Package version: 1.19.0

#### Steps to reproduce
```
$ python
Python 3.9.7 (default, Sep  3 2021, 06:18:44) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import proto
>>> class Blurb(proto.Message):
...   name = proto.Field(proto.STRING, number=1)
...   image = proto.Field(proto.BYTES, number=2)
... 
>>> d1 = {'name': 'ken', 'image': b'abc'}
>>> b1 = Blurb(d1)
>>> b1
name: "ken"
image: "abc"

>>> d2 = Blurb.to_dict(b1)
>>> d2
{'name': 'ken', 'image': 'YWJj'}
>>> b2 = Blurb(d2)
  File "<stdin>", line 1, in <module>
  File "/usr/local/google/home/kbandes/.local/lib/python3.9/site-packages/proto/message.py", line 519, in __init__
    super().__setattr__("_pb", self._meta.pb(**params))
TypeError: 'YWJj' has type str, but expected one of: bytes
>>> j1 = Blurb.to_json(b1)
>>> j1
'{\n  "name": "ken",\n  "image": "YWJj"\n}'
>>> b3 = Blurb.from_json(j1)
>>> b3
name: "ken"
image: "abc"
```
Note that to/from json works fine.

software-dov:
This is necessary for maintaining field integrity during round trip
conversions to python dicts.

E.g.

    class MyMessage(proto.Message):
          data = proto.Field(proto.BYTES, number=1)

    my_message = MyMessage(data=b"this is a data payload")
    assert my_message == MyMessage(MyMessage.to_dict(my_message))

Conversion to and from json does not exhibit this problem because the
entire ser/des logic is encapuslated by the vanilla protobuf runtime,
which handles this problem.

Fix for #249

### DIFF
--- a/docs/marshal.rst
+++ b/docs/marshal.rst
@@ -50,7 +50,7 @@ Protocol buffer type                Python type             Nullable
    This is necessary to preserve the original protobuf semantics when converting between
    Python dicts and proto messages.
    Converting a message containing a bytes field to a dict will
-   base64 encode the bytes field.
+   base64 encode the bytes field and yield a value of type str.
 
 .. code-block:: python
 
@@ -63,12 +63,14 @@ Protocol buffer type                Python type             Nullable
   msg = MyMessage(data=b"this is a message")
   msg_dict = MyMessage.to_dict(msg)
 
-  assert msg_dict == {'data': 'dGhpcyBpcyBhIG1lc3NhZ2U='}
+  # Note: the value is the base64 encoded string of the bytes field.
+  # It has a type of str, NOT bytes.
+  assert type(msg_dict['data']) == str
 
   msg_pb = ParseDict(msg_dict, MyMessage.pb())
-  msg = MyMessage(msg_dict)
+  msg_two = MyMessage(msg_dict)
 
-  assert msg_pb == msg
+  assert msg == msg_pb == msg_two
 
 
 Wrapper types
--- a/docs/marshal.rst
+++ b/docs/marshal.rst
@@ -44,6 +44,32 @@ Protocol buffer type                Python type             Nullable
     If you *write* a timestamp field using a Python ``datetime`` value,
     any existing nanosecond precision will be overwritten.
 
+.. note::
+
+   Setting a ``bytes`` field from a string value will first base64 decode the string.
+   This is necessary to preserve the original protobuf semantics when converting between
+   Python dicts and proto messages.
+   Converting a message containing a bytes field to a dict will
+   base64 encode the bytes field.
+
+.. code-block:: python
+
+  import proto
+  from google.protobuf.json_format import ParseDict
+
+  class MyMessage(proto.Message):
+      data = proto.Field(proto.BYTES, number=1)
+
+  msg = MyMessage(data=b"this is a message")
+  msg_dict = MyMessage.to_dict(msg)
+
+  assert msg_dict == {'data': 'dGhpcyBpcyBhIG1lc3NhZ2U='}
+
+  msg_pb = ParseDict(msg_dict, MyMessage.pb())
+  msg = MyMessage(msg_dict)
+
+  assert msg_pb == msg
+
 
 Wrapper types
 -------------
--- a/proto/fields.py
+++ b/proto/fields.py
@@ -131,6 +131,13 @@ class Field:
         if self.enum:
             return self.enum
 
+        # Special case for 'bytes'.
+        # We need to know if a field stores bytes
+        # so that we can properly url encode/decode strings.
+        # See marhsal.rules.bytes.BytesRule for more info.
+        if self.proto_type == ProtoType.BYTES:
+            return self.proto_type
+
         # For non-enum primitives, return None.
         if not self.message:
             return None
--- a/proto/fields.py
+++ b/proto/fields.py
@@ -134,7 +134,7 @@ class Field:
         # Special case for 'bytes'.
         # We need to know if a field stores bytes
         # so that we can properly url encode/decode strings.
-        # See marhsal.rules.bytes.BytesRule for more info.
+        # See marshal.rules.bytes.BytesRule for more info.
         if self.proto_type == ProtoType.BYTES:
             return self.proto_type
 
--- a/proto/fields.py
+++ b/proto/fields.py
@@ -126,21 +126,15 @@ class Field:
 
     @property
     def pb_type(self):
-        """Return the composite type of the field, or None for primitives."""
+        """Return the composite type of the field, or the primitive type if a primitive."""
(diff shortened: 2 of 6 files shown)


proto/fields.py
```diff
@@ -131,6 +131,13 @@ def pb_type(self):
         if self.enum:
             return self.enum
 
+        # Special case for 'bytes'.
+        # We need to know if a field stores bytes
+        # so that we can properly url encode/decode strings.
+        # See marhsal.rules.bytes.BytesRule for more info.
```
kbandes, line 137: type: marhsal -> marshal

proto/marshal/rules/bytes.py
```diff
@@ -0,0 +1,44 @@
...
+    Note: this conversion is asymmetric becasue Python does have a bytes type.
+    It is sometimes necessary to convert proto bytes fields to strings, e.g. for
+    JSON encoding, marshalling a message to a dict. Because bytes fields can
+    represent arbitrary data, bytes fields are base64 encoded when they need to
+    be represented as strings.
+
+    It is necessary to have the conversion be bidirectional, i.e.
+    my_message == MyMessage(MyMessage.to_dict(my_message))
+
+    To accomplish this, we need to intercept assignments from strigs and
```
kbandes, line 31: typo: strigs -> strings

tests/test_fields_bytes.py
```diff
@@ -71,5 +71,23 @@ class Foo(proto.Message):
     # for strings (but not vice versa).
     foo.bar = b"anything"
     assert foo.bar == "anything"
-    with pytest.raises(TypeError):
-        foo.baz = "anything"
+
+    # We need to permit setting bytes fields from strings,
+    # but the marhalling needs to base64 decode the result.
```
kbandes, line 76: typo: marhalling -> marshalling

proto/marshal/rules/bytes.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+
+    def to_python(self, value, *, absent: bool = None):
+        return value
+
+    def to_proto(self, value):
+        if isinstance(value, str):
+            value = value.encode("utf-8")
+            value += b"=" * (4 - len(value) % 4)  # padding
+            value = base64.urlsafe_b64decode(value)
```
tseaver, line 42: Why does this decode using the `urlsafe` alphabet?  Are there constraints on what might appear in `value` that aren't obvious here?

software-dov: This is what the vanilla proto runtime does when marshalling between JSON and proto. I just copied it over to maintain equality checking.

```python
import proto
from google.protobuf.json_format import ParseDict

class MyMessage(proto.Message):
    data = proto.Field(proto.BYTES, number=1)

msg = MyMessage(data=b"this is a message")
msg_dict = MyMessage.to_dict(msg)
msg = MyMessage(msg_dict)  # To trigger decoding

msg_pb = ParseDict(msg_dict, MyMessage.pb())

assert msg_pb == msg
```

proto/fields.py
software-dov: Good eye!

docs/marshal.rst
```diff
@@ -44,6 +44,32 @@ Protocol buffer type                Python type             Nullable
...
+  import proto
+  from google.protobuf.json_format import ParseDict
+
+  class MyMessage(proto.Message):
+      data = proto.Field(proto.BYTES, number=1)
+
+  msg = MyMessage(data=b"this is a message")
+  msg_dict = MyMessage.to_dict(msg)
+
+  assert msg_dict == {'data': 'dGhpcyBpcyBhIG1lc3NhZ2U='}
```
tseaver, line 66: That assertion conveys no value to the user -- better would be to use a constant for the un-encoded string value above, and assert that it is encoded here, e.g.:
```python
  assert msg_dict == {'data': base64.urlsafe_b64encode(message_data})
```

tests/test_fields_bytes.py
```diff
@@ -71,5 +71,23 @@ class Foo(proto.Message):
...
-    with pytest.raises(TypeError):
-        foo.baz = "anything"
+
+    # We need to permit setting bytes fields from strings,
+    # but the marshalling needs to base64 decode the result.
+    # This is a requirement for interop with the vanilla protobuf runtime:
+    # converting a proto message to a dict base64 encodes the bytes
+    # becase it may be sent over the network via a protocol like HTTP.
+    foo.baz = "dW5sYWRlbiBzd2FsbG93"
+    assert foo.baz == b"unladen swallow"
```
tseaver, line 81: Using the encoded literal is like having "magic numbers" in tests.  Please encode the value from a constant, e.g.:
```python
    unladen_swallow = b"unladen swallow"
    encoded_swallow = base64.urlsafe_b64encode(unladen_swallow)
    foo.baz = encoded_swallow
    assert foo.baz == unladen_swallow
```

tseaver (changes requested):

tests/test_fields_bytes.py
tswast: As with my other comment, this test as-is is much easier to verify than the proposed.

tswast (changes requested):

tests/test_fields_bytes.py
tseaver: Really?  You can eyeball that encoded string and see that it should decode to `b'unladen swallow'`?

proto/fields.py
```diff
@@ -131,6 +131,13 @@ def pb_type(self):
...
             return self.enum
 
+        # Special case for 'bytes'.
+        # We need to know if a field stores bytes
+        # so that we can properly url encode/decode strings.
+        # See marshal.rules.bytes.BytesRule for more info.
+        if self.proto_type == ProtoType.BYTES:
+            return self.proto_type
+
         # For non-enum primitives, return None.
```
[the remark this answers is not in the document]
software-dov: Yes, we consider `bytes` to be a primitive type. Anything that is not a message or an enum is a primitive.

tseaver (approved):

tswast (approved):

PR #270, opened 2021-10-26
test: verify json_name is respected

### LINKED ISSUE #130: json_name is not respected for proto.Field objects using Python proto impl
E.g.

```python
import os
os.environ["PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION"] = "python"

import proto
class Squid(proto.Message):
    massKg = proto.Field(proto.INT32, number=1, json_name="mass_in_kilograms")

s = Squid(massKg=20)
j = Squid.to_json(s)
print(j)    # '{"massKg" : 20}', or something like that.
```

This is only reproducible with the python protobuf implementation.

Editorial: my gut feeling is that this is a bug in proto-plus, or at least that a workaround/fix/correct solution is entirely possible within proto-plus.

software-dov:
Fields can customize their json name representation.
Add a test to verify that this customization is respected by both
Python and C++ runtimes.
Boost protobuf requirement to gain required upstream bugfix.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,7 @@ setup(
     long_description=README,
     platforms="Posix; MacOS X",
     include_package_data=True,
-    install_requires=("protobuf >= 3.12.0",),
+    install_requires=("protobuf >= 3.19.0",),
     extras_require={"testing": ["google-api-core[grpc] >= 1.22.2",],},
     python_requires=">=3.6",
     classifiers=[
--- a/noxfile.py
+++ b/noxfile.py
@@ -22,7 +22,16 @@ import nox
 CURRENT_DIRECTORY = pathlib.Path(__file__).parent.absolute()
 
 
-@nox.session(python=["3.6", "3.7", "3.8", "3.9"])
+PYTHON_VERSIONS = [
+    "3.6",
+    "3.7",
+    "3.8",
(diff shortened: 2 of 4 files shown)


parthea (approved):

setup.py
```diff
@@ -36,7 +36,7 @@
     long_description=README,
     platforms="Posix; MacOS X",
     include_package_data=True,
-    install_requires=("protobuf >= 3.12.0",),
+    install_requires=("protobuf >= 3.19.0",),
```
tswast, line 39: This should not have been included in a patch release. It's causing problems over in the conda ecosystem, where the version of `libprotobuf` is a bit slower to update as it require recompilation of many many packages. Currently the latest grcp-cpp package is compiled with `libprotobuf==3.18.1`.

tswast: Update: literally today is when they are rolling out changes to the build environment to support protobuf 3.19 https://github.com/conda-forge/grpc-cpp-feedstock/pull/124

tswast: For future reference, this is the where pin for conda was updated: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/2049

I recommend we don't update minimum protobuf beyond this in future

PR #280, opened 2021-12-08
fix: add pickling support to proto messages

### LINKED ISSUE #162: Proto Plus messages are not pickleable
From https://github.com/googleapis/python-iot/issues/42

```py
class Composer(proto.Message):
    given_name = proto.Field(proto.STRING, number=1)
    family_name = proto.Field(proto.STRING, number=2)

composer = Composer(given_name="Johannes", family_name="Bach")
pickled_composer = pickle.dumps(composer)
```

```
  File "<stdin>", line 1, in <module>
_pickle.PicklingError: Can't pickle <class 'Composer'>: it's not the same object as __main__.Composer
```

The current recommendation is to serialize to bytes. https://proto-plus-python.readthedocs.io/en/latest/reference/message.html

tseaver:
FBO of frameworks such as Apache Beam, which use them for sharing
state between trusted hosts.

Closes #260.

### DIFF
--- a/docs/messages.rst
+++ b/docs/messages.rst
@@ -246,8 +246,11 @@ already allows construction from mapping types.
 
 .. note::
 
-   Protobuf messages **CANNOT** be safely pickled or unpickled. This has serious consequences for programs that use multiprocessing or write messages to files.
-   The preferred mechanism for serializing proto messages is :meth:`~.Message.serialize`.
+   Although Python's pickling protocol has known issues when used with
+   untrusted collaborators, some frameworks do use it for communication
+   between trusted hosts.  To support such frameworks, protobuf messages
+   **can** be pickled and unpickled, although the preferred mechanism for
+   serializing proto messages is :meth:`~.Message.serialize`.
 
    Multiprocessing example:
 
--- a/proto/message.py
+++ b/proto/message.py
@@ -642,6 +642,16 @@ class Message(metaclass=MessageMeta):
         if pb_value is not None:
             self._pb.MergeFrom(self._meta.pb(**{key: pb_value}))
 
+    def __getstate__(self):
+        """Serialize for pickling."""
+        return self._pb.SerializeToString()
+
+    def __setstate__(self, value):
+        """Deserialization for pickling."""
+        new_pb = self._meta.pb().FromString(value)
+        super().__setattr__("_pb", new_pb)
+
+
 
 class _MessageInfo:
     """Metadata about a message.
--- a/proto/message.py
+++ b/proto/message.py
@@ -652,7 +652,6 @@ class Message(metaclass=MessageMeta):
         super().__setattr__("_pb", new_pb)
 
 
-
 class _MessageInfo:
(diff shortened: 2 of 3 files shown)


tswast:
I don't know enough about the details of pickling or what the "serious consequences" were when objects were pickled before, but this would greatly simplify implementations of connectors in tools like Dask and Beam.

Connectors such as https://beam.apache.org/releases/pydoc/2.33.0/apache_beam.io.gcp.bigtableio.html#apache_beam.io.gcp.bigtableio.WriteToBigTable are already pickling objects that contain proto-plus properties, so whatever "serious consequences" we are worried about are already happening over there.

tseaver:
@tswast There is a lot of fear in the Python community around pickles, because unpickling an object can import code:  if an attacker can get hostile (or compromised) code onto the system, and cause a process to unpickle an object which loads a class from the "evil" code, Bad Things(TM) happen.  So, loading pickles from untrusted sources can widen an attack surface.  Dask and Beam connectors don't really fall into that scenario, IMNSHO.

software-dov (approved):

software-dov:
I'm not very familiar with Beam, and it could be that this falls outside the scope of Beam's use case, but one additional complication is compatibility between different versions of the same message, e.g.
```python
# Old software version
from mollusc_v1 import Squid as SquidV1

s_v1 = SquidV1(mass_kg=10, length_cm=20)
s_pickle = pickle.dumps(s_v1)
with open("/tmp/squid.txt", "w") as f:
    f.write(s_pickle)

# Updated software version
from mollusc_v2 import Squid as SquidV2

with open("/tmp/squid.txt", "w") as f:
    s = f.read()

s_v2 = pickle.loads(s)

assert isinstance(s_v2, SquidV2)
```

What happens in the above scenario? Does the unpickling fail because the exact class isn't available? Is pickle smart/flexible enough to detect that SquidV2 is the class we have in mind? Are there subtle errors? Is the message class description serialized and deserialized as part of the pickle protocol?

tseaver:
@software-dov 
> Does the unpickling fail because the exact class isn't available? Is pickle smart/flexible enough to detect that SquidV2 is the class we have in mind? Are there subtle errors? Is the message class description serialized and deserialized as part of the pickle protocol?

A pickle of a class instance contains the dotted-importable name of the instance's type:  it ignores any `import foo as bar` aliasing.  The class is found via the unpickler's [`find_class`](https://docs.python.org/3/library/pickle.html#pickle.Unpickler.find_class) method.  

Another consequence is that instances of "local" classes (defined inside another function / method) aren't picklable, because their class can't be imported:  that is why I had to define the `Squid` class in `tests/test_message_pickling.py` outside of the method.

software-dov:
@tseaver Okay, that makes sense, and that answers my questions about old/new message interchange. The importable path will be the same assuming no breaking changes. It is then protobuf's responsibility to handle unknown/missing fields, which is well tested.

This does imply that it is *not* possible to pickle messages across a major version update.

PR #283, opened 2022-01-06
chore: add classifier for python 3.10

anukaal:
added support for python 3.10

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,7 +46,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: [3.6, 3.7, 3.8, 3.9, 3.10]
+        python: ['3.6', '3.7', '3.8', '3.9', '3.10']
         variant: ['', cpp]
     steps:
       - name: Cancel Previous Runs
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -46,7 +46,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: [3.6, 3.7, 3.8, 3.9]
+        python: [3.6, 3.7, 3.8, 3.9, 3.10]
```
software-dov, line 49: The really horrible annoying fact is that github actions doesn't like `3.10`: it needs to be in quotes.

software-dov:
Hi, thanks for contributing! Be sure to tweak the title of your PR to use [conventional commits](https://www.conventionalcommits.org/) before submitting.

.github/workflows/tests.yml
anukaal: > The really horrible annoying fact is that github actions doesn't like `3.10`: it needs to be in quotes.

I guess all the version should be in quotes that would be better

software-dov: That sounds good for consistency. My guess is that they're being interpreted as floating point numbers, which means stripping trailing zeros.

anukaal: Yeah , it would be great if all the python version should be in quotes

parthea (approved):

PR #289, opened 2022-02-04
feat: add custom __dir__ for messages and message classes

software-dov:
During development, it can be convenient to inspect objects and types
directly to determine what methods and attributes they have using the
dir() builtin command in a debugger or a REPL.

Because proto-plus messages wrap their fields using __getattr__, the
proto fields are not visible by default and must be explicitly exposed
to dir().

### DIFF
--- a/proto/message.py
+++ b/proto/message.py
@@ -273,6 +273,27 @@ class MessageMeta(type):
     def meta(cls):
         return cls._meta
 
+    def __dir__(self):
+        names = set(dir(type))
+        names.update(
+            (
+                "meta",
+                "pb",
+                "wrap",
+                "serialize",
+                "deserialize",
+                "to_json",
+                "from_json",
+                "to_dict",
+                "copy_from",
+            )
+        )
+        desc = self.pb().DESCRIPTOR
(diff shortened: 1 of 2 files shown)


proto/message.py
```diff
@@ -520,6 +541,29 @@ def __init__(
         # Create the internal protocol buffer.
         super().__setattr__("_pb", self._meta.pb(**params))
 
+    def __dir__(self):
+        desc = type(self).pb().DESCRIPTOR
+        names = {f_name for f_name in self._meta.fields.keys()}
+        names.update(m.name for m in desc.nested_types)
+        names.update(e.name for e in desc.enum_types)
+        names.update(dir(object()))
```
busunkim96, line 549: What does this do?

busunkim96 (approved):

proto/message.py
software-dov: It's including all the fields message fields, nested message and enum types, and boilerplate methods and attributes inherited from `object`. The important one is fields, since they weren't visible before, but since this is customizing `__dir__` we need to explicitly add all the others back in too.

PR #291, opened 2022-02-07
test: workaround for buggy pytest

software-dov:


parthea (approved):

mgorny:
Can this be reverted now that you've found the root cause, as indicated in pytest-dev/pytest#9637?

PR #301, opened 2022-02-16
fix: additional logic to mitigate collisions with reserved terms

parthea:
Additional fixes to mitigate collisions following PR #295

### DIFF
--- a/proto/message.py
+++ b/proto/message.py
@@ -699,18 +699,14 @@ class Message(metaclass=MessageMeta):
             their Python equivalents. See the ``marshal`` module for
             more details.
         """
-        try:
-            (key, pb_type) = self._get_pb_type_from_key(key)
-            if pb_type is None:
-                raise AttributeError(
-                    "Unknown field for {}: {}".format(self.__class__.__name__, key)
-                )
-
-            pb_value = getattr(self._pb, key)
-            marshal = self._meta.marshal
-            return marshal.to_python(pb_type, pb_value, absent=key not in self)
-        except KeyError as ex:
-            raise AttributeError(str(ex))
+        (key, pb_type) = self._get_pb_type_from_key(key)
+        if pb_type is None:
+            raise AttributeError(
+                "Unknown field for {}: {}".format(self.__class__.__name__, key)
+            )
+        pb_value = getattr(self._pb, key)
+        marshal = self._meta.marshal
+        return marshal.to_python(pb_type, pb_value, absent=key not in self)
 
     def __ne__(self, other):
         """Return True if the messages are unequal, False otherwise."""
--- a/proto/message.py
+++ b/proto/message.py
@@ -524,26 +524,14 @@ class Message(metaclass=MessageMeta):
         # coerced.
         marshal = self._meta.marshal
         for key, value in mapping.items():
-            try:
-                pb_type = self._meta.fields[key].pb_type
-            except KeyError:
-                # Underscores may be appended to field names
-                # that collide with python or proto-plus keywords.
-                # In case a key only exists with a `_` suffix, coerce the key
-                # to include the `_` suffix. It's not possible to
-                # natively define the same field with a trailing underscore in protobuf.
-                # See related issue
-                # https://github.com/googleapis/python-api-core/issues/227
-                if f"{key}_" in self._meta.fields:
-                    key = f"{key}_"
-                    pb_type = self._meta.fields[key].pb_type
-                else:
-                    if ignore_unknown_fields:
-                        continue
+            (key, pb_type) = self._get_pb_type_from_key(key)
+            if pb_type is None:
+                if ignore_unknown_fields:
+                    continue
 
-                    raise ValueError(
-                        "Unknown field for {}: {}".format(self.__class__.__name__, key)
-                    )
+                raise ValueError(
+                    "Unknown field for {}: {}".format(self.__class__.__name__, key)
+                )
 
             try:
                 pb_value = marshal.to_proto(pb_type, value)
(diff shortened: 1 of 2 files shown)


proto/message.py
```diff
@@ -545,7 +545,27 @@ def __init__(
...
+                # See related issue
+                # https://github.com/googleapis/python-api-core/issues/227
+                if isinstance(value, dict):
+                    keys_to_update = []
+                    for item in value:
+                        if not hasattr(pb_type, item) and hasattr(pb_type, f"{item}_"):
+                            keys_to_update.append(item)
+                    for item in keys_to_update:
+                        value[f"{item}_"] = value[item]
+                        del value[item]
```
software-dov, line 565: Outrageous code golf nit:
```python
value[f"{item}_"] = value.pop(item)
```

proto/message.py
```diff
@@ -662,7 +682,24 @@ def __getattr__(self, key):
...
+                # natively define the same field with a trailing underscore in protobuf.
+                # See related issue
+                # https://github.com/googleapis/python-api-core/issues/227
+                if f"{key}_" in self._meta.fields:
+                    key = f"{key}_"
+                    pb_type = self._meta.fields[key].pb_type
+                else:
+                    raise KeyError(
+                        "Unknown field for {}: {}".format(self.__class__.__name__, key)
+                    )
```
software-dov, line 701: I think this should actually be an `AttributeError` for consistency with the exception below.

proto/message.py
```diff
@@ -685,7 +722,24 @@ def __setattr__(self, key, value):
...
+            # natively define the same field with a trailing underscore in protobuf.
+            # See related issue
+            # https://github.com/googleapis/python-api-core/issues/227
+            if f"{key}_" in self._meta.fields:
+                key = f"{key}_"
+                pb_type = self._meta.fields[key].pb_type
+            else:
+                raise KeyError(
+                    "Unknown field for {}: {}".format(self.__class__.__name__, key)
+                )
```
software-dov, line 741: I think we should change this to an `AttributeError` for consistency with `__getattr__`.

tests/test_fields_mitigate_collision.py
```diff
@@ -27,10 +27,40 @@ class TestMessage(proto.Message):
...
     # Test that `spam` is coerced to `spam_`
     modified_obj = TestMessage({"spam": "has_spam", "eggs": "has_eggs"})
     assert modified_obj.spam_ == "has_spam"
+
+    # Test __setattr__ and __getattr___
+    modified_obj.__setattr__("spam", "no_spam")
+    modified_obj.__getattr__("spam") == "no_spam"
+
+    modified_obj.__setattr__("spam_", "yes_spam")
+    modified_obj.__getattr__("spam_") == "yes_spam"
```
software-dov, line 46: This works, but it's more idiomatic and arguably a better test to just try to interact with the attributes directly:
```python
modified_obj.spam = "no_spam"
assert modified_obj.spam == "no_spam"

modified_obj.spam_ = "yes_spam"
assert modified_obj.spam_ == "yes_spam"
```
It may be a good idea to go bidirectionally too:
```python
modified_obj.spam = "maybe_spam"
assert modified_obj.spam_ == "maybe_spam"

modified_obj.spam_ = "maybe_not_spam"
assert modified_obj.spam == "maybe_not_spam"
```

tests/test_fields_mitigate_collision.py
```diff
@@ -27,10 +27,40 @@ class TestMessage(proto.Message):
...
+        text_stream=TestMessage({"spam": "has_spam", "eggs": "has_eggs"})
+    )
+    assert modified_obj.text_stream.spam_ == "has_spam"
+
+    # Test __setattr__ and __getattr___
+    modified_obj.text_stream.__setattr__("spam", "no_spam")
+    modified_obj.text_stream.__getattr__("spam") == "no_spam"
+
+    modified_obj.text_stream.__setattr__("spam_", "yes_spam")
+    modified_obj.text_stream.__getattr__("spam_") == "yes_spam"
```
software-dov, line 59: Same here as above: use the attributes directly, cross test.

tests/test_fields_mitigate_collision.py
```diff
@@ -27,10 +27,40 @@ class TestMessage(proto.Message):
...
+
+    with pytest.raises(KeyError):
+        modified_obj.text_stream.__setattr__("key_does_not_exist", "n/a")
```
software-dov, line 62: Same here: use the attribute directly.
We should also add a test for `__getattr__`

proto/message.py
```diff
@@ -685,7 +722,24 @@ def __setattr__(self, key, value):
         if key[0] == "_":
             return super().__setattr__(key, value)
         marshal = self._meta.marshal
-        pb_type = self._meta.fields[key].pb_type
+        try:
+            pb_type = self._meta.fields[key].pb_type
+        except KeyError:
```
vam-google, line 727: These three newly introduced sections seem very similar (and the comment section too). Can we put it in a helper method or something to avoid code duplication? The small differences can be addressed by passing some sort of parameters into a helper funciton?

vam-google (commented):
Look goo to me, but can we please avoid code dupliction if possible?

tests/test_fields_mitigate_collision.py
```diff
@@ -38,28 +38,43 @@ class TextStream(proto.Message):
...
+    modified_obj.text_stream.spam = "maybe_spam"
+    assert modified_obj.text_stream.spam_ == "maybe_spam"
+
+    modified_obj.text_stream.spam_ = "maybe_not_spam"
+    assert modified_obj.text_stream.spam == "maybe_not_spam"
 
-    modified_obj.text_stream.__setattr__("spam_", "yes_spam")
-    modified_obj.text_stream.__getattr__("spam_") == "yes_spam"
+    with pytest.raises(AttributeError):
+        modified_obj.text_stream.attribute_does_not_exist == "n/a"
```
software-dov, line 74: Nit: should add an `assert` right at the beginning of the line.

software-dov (approved):
One minor nit, otherwise LGTM.

tests/test_fields_mitigate_collision.py
parthea: Done

PR #176, opened 2021-01-08
fix: default proto package name is the module name, not ""

0x2b3bfa0:
Closes #175; please refer to #171 for context.

PR #309, opened 2022-04-19
fix: default proto package name is the module name, not ""

### LINKED ISSUE #175: fix: default proto package name is the module name, not ""
See #171 for background information.

parthea:
Fixes #175

This is similar to #176 but the PR is stale and the branch is no longer active


vam-google (approved):
LGTM

PR #312, opened 2022-04-29
ci: refactor coverage check

parthea:
This PR refactors the github action coverage check to test combined coverage across all checks. This is needed for https://github.com/googleapis/proto-plus-python/pull/226 as there is a coverage gap when checks are evaluated individually rather than combined.

We use a similar check in googleapis python-* repos. See the templated file [here](https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_library/.github/workflows/unittest.yml#L45).

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,7 +2,6 @@
 branch = True
 
 [report]
-fail_under = 100
 show_missing = True
 omit =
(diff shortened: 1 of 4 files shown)


vam-google (approved):
LGTM

PR #327, opened 2022-06-24
feat: add initial files to allow protobuf 4.20

atulep:
BEGIN_COMMIT_OVERRIDE
feat: add initial files to allow protobuf 4.20
END_COMMIT_OVERRIDE

This PR integrates protobuf 4.xx into codebase.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -42,7 +42,8 @@ def unit(session, proto="python"):
     session.env["PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION"] = proto
     session.install("coverage", "pytest", "pytest-cov", "pytz")
     session.install("-e", ".[testing]", "-c", constraints_path)
-
+    if proto == 'cpp': # 4.20 does not have cpp.
+        session.install("protobuf==3.19.0")
     session.run(
         "py.test",
         "-W=error",
--- a/noxfile.py
+++ b/noxfile.py
@@ -42,7 +42,7 @@ def unit(session, proto="python"):
     session.env["PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION"] = proto
     session.install("coverage", "pytest", "pytest-cov", "pytz")
     session.install("-e", ".[testing]", "-c", constraints_path)
-    if proto == 'cpp': # 4.20 does not have cpp.
+    if proto == "cpp":  # 4.20 does not have cpp.
         session.install("protobuf==3.19.0")
     session.run(
         "py.test",
--- a/noxfile.py
+++ b/noxfile.py
@@ -54,7 +54,7 @@ def unit(session, proto="python"):
                 "--cov-config=.coveragerc",
                 "--cov-report=term",
                 "--cov-report=html",
-                os.path.join("tests", ""),
+                os.path.join("tests"),
             ]
         ),
     )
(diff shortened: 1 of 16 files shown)


noxfile.py
```diff
@@ -54,7 +55,7 @@ def unit(session, proto="python"):
                 "--cov-config=.coveragerc",
                 "--cov-report=term",
                 "--cov-report=html",
-                os.path.join("tests", ""),
+                os.path.join("tests"),
```
parthea, line 58:
```suggestion
                "tests",
```

proto/marshal/collections/maps.py
```diff
@@ -58,15 +59,20 @@ def __getitem__(self, key):
...
-
         # Directly setting a key is not allowed; however, protocol buffers
         # is so permissive that querying for the existence of a key will in
         # of itself create it.
         #
         # Therefore, we create a key that way (clearing any fields that may
         # be set) and then merge in our values.
-        self.pb[key].Clear()
-        self.pb[key].MergeFrom(pb_value)
+        # In UPB, sometime self.pb[key] is not always a proto.
```
parthea, line 68:
```suggestion
        # In UPB, sometimes self.pb[key] is not always a proto.
```

parthea (approved):

proto/_package_info.py
```diff
@@ -37,7 +37,12 @@ def compile(name, attrs):
     proto_module = getattr(module, "__protobuf__", object())
 
     # A package should be present; get the marshal from there.
-    package = getattr(proto_module, "package", module_name)
+    # When package is empty, upb based protobuf fails with an
```
vam-google, line 40: Please mark all hacks like this with a TODO saying it should be removed once the new version of protobuf with fixed upb is out.

proto/marshal/compat.py
```diff
@@ -25,9 +25,17 @@
 except ImportError:
     _message = None
 
+# Try UPB.
+if not _message:
```
vam-google, line 29: Please test it more as we discussed. In case if ptotobuf 3.20 and 4.21 re mutually exclusive in this logic, please prioritize workflow, when newer version of protobuf if assumed first, and only if it fails, fallback to older version, not vice versa. This is to make sure that throwing an exception is not an expected behavior for the prioritized protobuf version (which I guess should be the new one)

proto/message.py
```diff
@@ -568,11 +573,21 @@ def __init__(
                 # See related issue
                 # https://github.com/googleapis/python-api-core/issues/227
                 if isinstance(value, dict):
-                    keys_to_update = [
-                        item
-                        for item in value
-                        if not hasattr(pb_type, item) and hasattr(pb_type, f"{item}_")
-                    ]
+                    if _message is None:
```
vam-google, line 576: For this code block (and other similar blocks in this PR) please add an explicit comment for each if branch which protobuf version it belongs to (3.20 or 4.21 case).

tests/conftest.py
```diff
@@ -37,10 +37,27 @@ def pytest_runtest_setup(item):
...
+
+            has_upb = True
+            assert _message
+        except ImportError:
+            _message = None
+        if not _message:
+            try:
+                from google.protobuf.pyext import _message
+            except ImportError:
+                _message = None
```
vam-google, line 52: Consider having a utility helper method or something similar for htis logic, as it seems it is being reused in multiple places

vam-google:
@haberman can you please take a look at this PR?

proto/_package_info.py
atulep: Done.

proto/message.py
atulep: Done

vam-google (approved):
LGTM

PR #332, opened 2022-07-26
fix: fixes bug in the test.

atulep:
Fixes failing test after new protobuf was released.

The `ScalarMap` is being returned and needs to be explicitly converted to `dict` for assertion to succeed.


vam-google (approved):
LGTM

PR #334, opened 2022-08-04
fix: Fix Timestamp, Duration and FieldMask marshaling in REST transport

### LINKED ISSUE #333: Proto JSON cannonical encoding is not supported for special types
With an exception of `int64<->string` encoding (implemented in https://github.com/googleapis/proto-plus-python/pull/267) the rest of special types [JSON Mapping ](https://developers.google.com/protocol-buffers/docs/proto3#json) are not supported. Most commonly failing types are `Timestamp`, `FieldMask` and `Duration`

This is crucial to support rest transport in Python.

vam-google:
This fixes https://github.com/googleapis/proto-plus-python/issues/333

BEGIN_COMMIT_OVERRIDE
fix: Fix Timestamp, Duration and FieldMask marshaling in REST transport
Release-As: 1.22.0
END_COMMIT_OVERRIDE

### DIFF
--- a/proto/marshal/rules/field_mask.py
+++ b/proto/marshal/rules/field_mask.py
@@ -14,6 +14,7 @@
 
 from google.protobuf import field_mask_pb2
 
+
 class FieldMaskRule:
     """A marshal between FieldMask and strings.
 
--- a/proto/marshal/rules/field_mask.py
+++ b/proto/marshal/rules/field_mask.py
@@ -1,0 +1,35 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from google.protobuf import field_mask_pb2
+
+class FieldMaskRule:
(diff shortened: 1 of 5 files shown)


proto/marshal/rules/field_mask.py
```diff
@@ -0,0 +1,36 @@
...
+class FieldMaskRule:
+    """A marshal between FieldMask and strings.
+
+    See https://github.com/googleapis/proto-plus-python/issues/333
+    and
+    https://developers.google.com/protocol-buffers/docs/proto3#json
+    for more details.
+    """
+
+    def to_python(self, value, *, absent: bool = None):
```
atulep, line 27: Is it expected to be a string ?

proto/marshal/rules/field_mask.py
```diff
@@ -0,0 +1,36 @@
...
+        return value
+
+    def to_proto(self, value):
```
atulep, line 30: What are possible types for value ?

atulep: Is this protoplus proto or protobuf proto ?

atulep (approved):

proto/marshal/rules/field_mask.py
atulep, line 27: This is not used - should be removed ?

vam-google: what specifically? This just follows a similar convention of other Rules (they do exaclty the same thing as I have here).

proto/marshal/rules/field_mask.py
vam-google: Proto-plus

proto/marshal/rules/field_mask.py
vam-google: It is not used, but other rules still have it (i ti snot used in most places), so keeping it like this for consistency.

PR #336, opened 2022-08-27
fix: Add no-pretty print option

### LINKED ISSUE #284: to_json option to exclude newline
The `to_json` result always pretty-prints, meaning newlines are always inserted. This makes the result unusable with apps expecting newline-delimited JSON, that is, strict json w/o formatting (no LF, tabs,...). 
Our use case is BigQuery which treats newlines as record-separator, and trying to read a `to-json` file or record returns `key not found`.

Request:
There should be an option to output the result without formatting, no newlines or tabs.

Workaround:
use `to_dict` and use json module to correctly convert.

vam-google:
This fixes https://github.com/googleapis/proto-plus-python/issues/284 and in general  makes sense.

### DIFF
--- a/proto/message.py
+++ b/proto/message.py
@@ -400,7 +400,7 @@ class MessageMeta(type):
             use_integers_for_enums=use_integers_for_enums,
             including_default_value_fields=including_default_value_fields,
             preserving_proto_field_name=preserving_proto_field_name,
-            indent=indent
+            indent=indent,
         )
 
     def from_json(cls, payload, *, ignore_unknown_fields=False) -> "Message":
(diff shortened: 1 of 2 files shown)


vchudnov-g (approved):

parthea (commented):
Please check if we can address new lines also: https://github.com/googleapis/proto-plus-python/issues/284#issuecomment-1229110388

vam-google:
@parthea The new lines have been addressed originally. I've responded to thebug and also improved the documentation to make it clear that `indent=None` means no new lines.

PR #341, opened 2022-09-29
fix: Fix the issue with empty package name

nodarai:
This subsequently fixes the issue with nested messages reported in googleapis/google-cloud-python#15127.


parthea:
Hi @nodarai , Thanks for the PR! We still need to support older versions of `protobuf`. Since we still need to support older versions of protobuf, perhaps we could use google.protobuf.__version__ to check whether the workaround is needed. Assigning @atulep to take a look as the author of that section.

nodarai:
@parthea @atulep What do you think we should change/add in order to proceed?

parthea:
Hi @atulep , please could you take another look?

parthea:
Converting to draft as https://github.com/googleapis/proto-plus-python/issues/389 needs to be completed first.

parthea:
Closing as stale. We will revisit issue https://github.com/googleapis/proto-plus-python/issues/340 once https://github.com/googleapis/proto-plus-python/issues/389 is addressed.

PR #213, opened 2021-03-12
feat: add preserving_proto_field_name to to_json

software-dov:

PR #384, opened 2023-08-03
feat: Add additional parameters to `to_json()` and `to_dict()` methods

holtskinner:
- These parameters are just passthroughs to [`MessageToJson`][1] and [`MessageToDict`][2]

[1]: https://googleapis.dev/python/protobuf/latest/google/protobuf/json_format.html#google.protobuf.json_format.MessageToJson

[2]: https://googleapis.dev/python/protobuf/latest/google/protobuf/json_format.html#google.protobuf.json_format.MessageToDict

### DIFF
--- a/proto/message.py
+++ b/proto/message.py
@@ -393,7 +393,7 @@ class MessageMeta(type):
                 proto case (snake_case) or use lowerCamelCase. Default is False.
             sort_keys (Optional(bool)): If True, then the output will be sorted by field names.
                 Default is False.
-            indent (int): The JSON object will be pretty-printed with this indent level.
+            indent (Optional(int)): The JSON object will be pretty-printed with this indent level.
                 An indent level of 0 or negative will only insert newlines.
                 Pass None for the most compact representation without newlines.
             float_precision (Optional(int)): If set, use this to specify float field valid digits.
--- a/proto/message.py
+++ b/proto/message.py
@@ -376,7 +376,10 @@ class MessageMeta(type):
         use_integers_for_enums=True,
         including_default_value_fields=True,
         preserving_proto_field_name=False,
+        sort_keys=False,
         indent=2,
+        float_precision=None,
+        ensure_ascii=False,
     ) -> str:
         """Given a message instance, serialize it to json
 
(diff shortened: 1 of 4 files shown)


holtskinner:
I think Owlbot has to be force run by adding the tag. I don't have permission to do that.

holtskinner:
Is this good for review?

proto/message.py
```diff
@@ -389,10 +391,13 @@ def to_json(
             preserving_proto_field_name (Optional(bool)): An option that
                 determines whether field name representations preserve
                 proto case (snake_case) or use lowerCamelCase. Default is False.
-            indent: The JSON object will be pretty-printed with this indent level.
+            sort_keys (Optional(bool)): If True, then the output will be sorted by field names.
+                Default is False.
+            indent (int): The JSON object will be pretty-printed with this indent level.
```
parthea, line 396:
```suggestion
            indent (Optional(int)): The JSON object will be pretty-printed with this indent level.
```

parthea (changes requested):
I apologize for the delay. The changes look good. Please can you also add tests, similar to  the ones added in https://github.com/googleapis/proto-plus-python/pull/164 and https://github.com/googleapis/proto-plus-python/pull/213

tests/test_message.py
```diff
@@ -271,6 +271,16 @@ class Color(proto.Enum):
...
 
 
+def test_serialize_to_dict_float_precision():
+    class Squid(proto.Message):
+        mass_kg = proto.Field(proto.FLOAT, number=1)
+
+    s = Squid(mass_kg=3.14159265)
+
+    s_dict = Squid.to_dict(s, float_precision=3)
+    assert s_dict["mass_kg"] == 3.14
```
parthea, line 281: I would have expected this to be `3.141` or `3.142` since `float_precision=3`. If you agree, can you file a bug?

holtskinner: Ok, I filed https://github.com/googleapis/proto-plus-python/issues/390

Should it block this PR?

parthea (approved):

PR #456, opened 2024-06-11
fix: drop python 3.6

parthea:
Remove python 3.6 which has already been dropped throughout the `googleapis` organization.

### DIFF
--- a/.github/sync-repo-settings.yaml
+++ b/.github/sync-repo-settings.yaml
@@ -7,9 +7,6 @@ branchProtectionRules:
   requiredStatusCheckContexts:
     - 'style-check'
     - 'docs'
-    - 'unit (3.6, cpp)'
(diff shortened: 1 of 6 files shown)


ohmayr:
Do we want to also include the following changes i.e.
- remove [testing/constraints-3.6.txt](https://github.com/googleapis/proto-plus-python/blob/0f893724cabe513a5a9f9c8428dbd31f1b4f1d52/testing/constraints-3.6.txt#L4)
- Update the following: https://github.com/googleapis/proto-plus-python/blob/0f893724cabe513a5a9f9c8428dbd31f1b4f1d52/.readthedocs.yml#L6

Looks good to me otherwise. Thanks for doing this @parthea!

ohmayr (approved):

PR #471, opened 2024-07-09
fix: Fixed the compatibility with conda, protobuf 4 and pyext protobuf

### LINKED ISSUE #470: proto-plus returns incorrect unwrapped protos when protobuf 4 is used in conda env
#### Environment details

  - Programming language: Python
  - OS: Linux
  - Language runtime version: 3.11, Anaconda
  - Package version: 1.24

#### Steps to reproduce

```
conda config --append channels conda-forge
conda create -n myenv
conda activate myenv 
pip install google-cloud-aiplatform 
```
```
$ python -c 'from vertexai.generative_models import _generative_models as g
part1 = g.Part.from_uri(uri="gs://foo/bar", mime_type="baz")
gapic_content = g._to_content(part1)
parts = gapic_content.parts
part0 = parts[0]
print(type(part0))
# print("text" in part0)
# print("file_data" in part0)
print(part0._pb.WhichOneof("data"))
```
```
<class 'Part'>
  File "<string>", line 9, in <module>
AttributeError: _pb
```

```
(issue_3129) user@w-avolkov-lydu7cih:~$ python
Python 3.11.0 (main, Mar  1 2023, 18:26:19) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from vertexai.generative_models import _generative_models as g
>>> part1 = g.Part.from_uri(uri="gs://foo/bar", mime_type="baz")
>>> gc = g._to_content(part1)

>>> type(gc)
<class 'google.cloud.aiplatform_v1beta1.types.content.Content'>
>>> type(gc.parts)
<class 'google.protobuf.pyext._message.RepeatedCompositeContainer'>
>>> type(gc.parts[0])
<class 'Part'>
>>> type(gc.parts[0]).__module__
>>> type(gc.parts[0]).__name__
'Part'
>>> type(gc.parts[0]).__dict__
mappingproxy({'DESCRIPTOR': <google.protobuf.pyext._message.MessageDescriptor object at 0x79c4ee245510>, '__module__': None, '__slots__': (), '__doc__': None, 'text': <field property 'google.cloud.aiplatform.v1beta1.Part.text'>, 'inline_data': <field property 'google.cloud.aiplatform.v1beta1.Part.inline_data'>, 'file_data': <field property 'google.cloud.aiplatform.v1beta1.Part.file_data'>, 'function_call': <field property 'google.cloud.aiplatform.v1beta1.Part.function_call'>, 'function_response': <field property 'google.cloud.aiplatform.v1beta1.Part.function_response'>, 'video_metadata': <field property 'google.cloud.aiplatform.v1beta1.Part.video_metadata'>})
```

See how `gc` is a proper GAPIC proto-plus message, but `gc.parts[0]` is a raw protobuf message.

See also https://github.com/googleapis/python-aiplatform/issues/3129

### LINKED ISSUE #3129: `TypeError: argument of type 'Part' is not iterable`
When we execute the code from the test playground for Gemini in Vertex AI Studio, there is a type error. Actually, `_raw_part` is not a dict. It is of type `Part`, as an extension of `proto.Message`. So, the `text` property of the class `Part` is not  available as keyword...

I have found no way to extract text from a streaming response other than using `repr` function and trimming it.

#### Environment details

  - OS type and version: `MacOS Sonoma 14.2`
  - Python version: `3.11.5`
  - pip version: `23.2.1`
  - `google-cloud-aiplatform` version:
```
Name: google-cloud-aiplatform
Version: 1.38.1
Summary: Vertex AI API client library
Home-page: https://github.com/googleapis/python-aiplatform
Author: Google LLC
Author-email: googleapis-packages@google.com
License: Apache 2.0
Location: /Users/xxxxxxx/anaconda3/lib/python3.11/site-packages
Requires: google-api-core, google-cloud-bigquery, google-cloud-resource-manager, google-cloud-storage, packaging, proto-plus, protobuf, shapely
Required-by: 
```

#### Steps to reproduce

  1. Just execute the example code below, after logging with `gcloud`

#### Code example

```python
import vertexai
from vertexai.preview.generative_models import GenerativeModel  # , Part


def generate():
    """Generates text using the Generative Model."""
    # Initialize Vertex AI
    vertexai.init(project="<project_id>", location="us-central1")

    model = GenerativeModel("gemini-pro")
    responses = model.generate_content(
        """hi. I would like to know the 10 first items of the fibonacci
        sequence.""",
        generation_config={
            "max_output_tokens": 2048,
            "temperature": 0.9,
            "top_p": 1
        },
        stream=True,
    )

    for i, response in enumerate(responses):
        print("Response:", i)
        print(response.candidates[0].content.parts)
        # print(repr(response.candidates[0].content.parts[0]).split("\"")[1])
        print(response.candidates[0].content.parts[0].text)

generate()
```

#### Stack trace
```
Response: 0
[text: "1. 0\n2. 1\n3. 1\n4"
]
  File "/Users/xxxxxxxxx/Documents/002-git/dolibarr-invoice/services.py", line 57, in <module>
    generate()
  File "/Users/xxxxxxxxx/Documents/002-git/dolibarr-invoice/services.py", line 54, in generate
    print(response.candidates[0].content.parts[0].text)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/xxxxxxxxx/anaconda3/lib/python3.11/site-packages/vertexai/generative_models/_generative_models.py", line 1508, in text
    if "text" not in self._raw_part:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'Part' is not iterable
```

Ark-kun:
Fixes: https://github.com/googleapis/proto-plus-python/issues/470
Fixes: https://github.com/googleapis/python-aiplatform/issues/3129


parthea (commented):
Is it possible to add a test that fails? I'm not able to reproduce the issue.

happy-qiao:
Could you elaborate how this change can fix pyext issue? Essentially, the PR doesn't seem to change the value of exported variables: "repeated_composite_types",  "repeated_scalar_types", "map_composite_types",

7hmJ6Rkgc2spo4n (1)

Ark-kun:
@parthea 

>Is it possible to add a test that fails? 

Do you have Conda tests environments?

>I'm not able to reproduce the issue.

I think you've added repro steps to the internal doc.

Also there are repro steps in the linked issue: https://github.com/googleapis/proto-plus-python/issues/470

Ark-kun:
>Could you elaborate how this change can fix pyext issue? Essentially, the PR doesn't seem to change the value of exported variables: "repeated_composite_types", "repeated_scalar_types", "map_composite_types".

The code does change the items of the lists and the number of those items.

>Is it for the case when upb and pyext both exists

Yes. This is the exact case where you and our customers have issues.

> and the change makes it default to pyext?

No. The change does not change the default. It **adds** the pyext types to the lists.

>But pyext is deprecated

pyext is used by real customers who want to use Vertex AI. pyext is used by default in protobuf installed by conda.

parthea (commented):
Please could you add a comment in the code which explains the reason that we need to include both `upb` and `pyext` types? You mentioned this in the PR comments but it should be included in code comments as well.

Please could you also fix the failing presubmits ? This PR needs to be updated to cater for https://github.com/googleapis/proto-plus-python/pull/474 which was merged

Otherwise, LGTM

parthea (commented):
Would you be able to refactor the code to reduce duplication? For example, can we use a helper function instead of having repeated code?

proto/marshal/compat.py
```diff
@@ -55,6 +49,25 @@
         pass
 
 
+
+try:
+    from google.protobuf.pyext import _message as _message_pyext
+except ImportError:
+    _message_pyext = None
```
parthea, line 56: Move this import statement to the top of the file since we're importing it as `_message_pyext`

PR #479, opened 2024-07-23
fix: construct messages with nested struct

ohmayr:
This PR fixes the issue to construct a message from a dictionary when a message has a nested struct.

```python3
from google.protobuf import struct_pb2
import proto


class Foo(proto.Message):
    struct_field: struct_pb2.Struct = proto.Field(
        proto.MESSAGE,
        number=1,
        message=struct_pb2.Struct,
    )

class Bar(proto.Message):
    foo_field: Msg1 = proto.Field(
        proto.MESSAGE,
        number=1,
        message=Msg1,
    )

# The following works:
Foo({"struct_field": {"foo": "cheese"}})

# The following does not work (without the fix):
Bar({"foo_field": {"struct_field": {"foo": "cheese"}}})

```

Googlers see b/321905145

### DIFF
--- a/proto/marshal/rules/message.py
+++ b/proto/marshal/rules/message.py
@@ -37,7 +37,7 @@ class MessageRule:
             except (TypeError, ValueError) as ex:
                 # If we have a TypeError or Valueerror,
                 # try the slow path in case the error
-                # was: 
+                # was:
                 # - an int64/string issue.
                 # - a missing key issue in case a key only exists with a `_` suffix.
                 #   See related issue: https://github.com/googleapis/python-api-core/issues/227.
--- a/proto/marshal/rules/message.py
+++ b/proto/marshal/rules/message.py
@@ -34,7 +34,8 @@ class MessageRule:
             try:
                 # Try the fast path first.
                 return self._descriptor(**value)
-            except TypeError as ex:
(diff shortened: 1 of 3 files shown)


parthea:
LGTM but holding off on formal review until `cover` is fixed

proto/marshal/rules/message.py
```diff
@@ -34,7 +34,7 @@ def to_proto(self, value):
             try:
                 # Try the fast path first.
                 return self._descriptor(**value)
-            except TypeError as ex:
+            except (TypeError, ValueError) as ex:
                 # If we have a type error,
```
parthea, line 38:
```suggestion
                # If we have TypeError or ValueError,
```

proto/marshal/rules/message.py
```diff
@@ -34,7 +34,7 @@ def to_proto(self, value):
             try:
                 # Try the fast path first.
                 return self._descriptor(**value)
-            except TypeError as ex:
+            except (TypeError, ValueError) as ex:
```
parthea, line 37: Please also add a link to a bug to add more context on why we need to handle ValueError

parthea (approved):

parthea (approved):

PR #492, opened 2024-10-15
fix: fix 'Couldn't build proto file' when using Python 3.13

### LINKED ISSUE #490: Couldn't build proto file into descriptor pool [Python 3.13]
Encountered this issue while testing `Python 3.13.0rc2`. To reproduce

```py
import proto

__protobuf__ = proto.module(
    package="some_module",
    manifest={"View", "Encoding"},
)

class View(proto.Enum):
    A = 0

class Encoding(proto.Enum):
    B = 0
```

```py
$ python test.py
  File "/.../proto-plus-python/test.py", line 11, in <module>
    class Encoding(proto.Enum):
        B = 0
  File "/.../proto-plus-python/proto/enums.py", line 105, in __new__
    file_info.generate_file_pb(new_class=cls, fallback_salt=full_name)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.../proto-plus-python/proto/_file_info.py", line 104, in generate_file_pb
    pool.Add(self.descriptor)
    ~~~~~~~~^^^^^^^^^^^^^^^^^
TypeError: Couldn't build proto file into descriptor pool: duplicate symbol 'some_module.__firstlineno__'
```

```
$ pip list
Package    Version
---------- -------
pip        24.2
proto-plus 1.24.0
protobuf   5.28.2
```

--
AFAICT the issue shows up as soon as two `proto.Enum` are defined in one module. A quick experiment suggests that the reason might be here
https://github.com/googleapis/proto-plus-python/blob/e5e85339ffce7660253a60065927cf791a4ed2b7/proto/enums.py#L72-L78
The attributes are only filtered by `if isinstance(number, int)`. Python 3.13 added the new `__firstlineno__` attribute which is an integer and as such not filtered as it should be. Just changing the condition here would be enough to resolve it, just not sure that's the best option as I'm unfamiliar with the code base.
```diff
-if isinstance(number, int)
+if isinstance(number, int) and name != "__firstlineno__"
```

parthea:
Fixes https://github.com/googleapis/proto-plus-python/issues/490

### DIFF
--- a/proto/enums.py
+++ b/proto/enums.py
@@ -74,7 +74,7 @@ class ProtoEnumMeta(enum.EnumMeta):
                     descriptor_pb2.EnumValueDescriptorProto(name=name, number=number)
                     # Minor hack to get all the enum variants out.
                     for name, number in attrs.items()
-                    if isinstance(number, int)
+                    if name in attrs._member_names and isinstance(number, int)
                 ),
                 key=lambda v: v.number,
             ),
--- a/proto/enums.py
+++ b/proto/enums.py
@@ -73,6 +73,8 @@ class ProtoEnumMeta(enum.EnumMeta):
                 (
                     descriptor_pb2.EnumValueDescriptorProto(name=name, number=number)
                     # Minor hack to get all the enum variants out.
+                    # Use the `_member_names` property to get only the enum members
+                    # See https://github.com/googleapis/proto-plus-python/issues/490
                     for name, number in attrs.items()
                     if name in attrs._member_names and isinstance(number, int)
                 ),
--- a/tests/clam.py
+++ b/tests/clam.py
@@ -19,6 +19,7 @@ __protobuf__ = proto.module(
     manifest={
         "Clam",
         "Species",
+        "Color",
     },
 )
 
@@ -30,6 +31,14 @@ class Species(proto.Enum):
     GIGAS = 3
 
 
+class Color(proto.Enum):
+    COLOR_UNKNOWN = 0
+    BLUE = 1
+    ORANGE = 2
+    GREEN = 3
+
+
 class Clam(proto.Message):
     species = proto.Field(proto.ENUM, number=1, enum="Species")
     mass_kg = proto.Field(proto.DOUBLE, number=2)
+    color = proto.Field(proto.ENUM, number=3, enum="Color")


tests/clam.py
```diff
@@ -30,6 +31,14 @@ class Species(proto.Enum):
     GIGAS = 3
 
 
+class Color(proto.Enum):
+    COLOR_UNKNOWN = 0
```
ohmayr, line 35:
```suggestion
    UNKNOWN = 0
```

proto/enums.py
```diff
@@ -74,7 +74,7 @@ def __new__(mcls, name, bases, attrs):
                     descriptor_pb2.EnumValueDescriptorProto(name=name, number=number)
                     # Minor hack to get all the enum variants out.
                     for name, number in attrs.items()
-                    if isinstance(number, int)
+                    if name in attrs._member_names and isinstance(number, int)
```
ohmayr, line 77: Can you explain why this check is needed?

ohmayr (approved):

tests/clam.py
parthea: `COLOR_UNKNOWN` follows protobuf best practices for the default value

`In proto3, the first value defined in an enum definition must have the value zero and should have the name ENUM_TYPE_NAME_UNSPECIFIED or ENUM_TYPE_NAME_UNKNOWN`

https://protobuf.dev/programming-guides/proto3/#enum-default

proto/enums.py
parthea: Fixed in https://github.com/googleapis/proto-plus-python/pull/492/commits/d72dfb431c6e529c234677c3d0fcf1ea6afc3404

`attrs.items()` contains more than just the members. `attrs._member_names` contains only the members

For example, see the result of `attrs.keys()`
```
attrs.keys()
dict_keys(['_generate_next_value_', '__module__', '__qualname__', '__firstlineno__', 'UNKNOWN', 'SQUAMOSA', 'DURASA', 'GIGAS', '__static_attributes__'])
```

tests/clam.py
ohmayr: Sounds good!

PR #502, opened 2024-11-01
fix: Use include rather than exclude to find_namespace_packages in setup.py

### LINKED ISSUE #503: running setup.py bdist_wheel results in build/, docs/ and testing/ in the wheel
Thanks for stopping by to let us know something could be better!

#### Environment details

  - Programming language: python
  - OS: ubuntu 24.04
  - Language runtime version: 3.12
  - Package version: git

#### Steps to reproduce

  1. Run `setup.py bdist_wheel` with pip-installed versions of setuptools, wheel
  2. look at created wheel, it will have testing/, docs/ and build/ directories in it.

Attached [do-build.sh.txt](https://github.com/user-attachments/files/17604156/do-build.sh.txt) will recreate.

Example trimmed outputt:
```
% sh  /tmp/do-build.sh.txt
$ mkdir -p /tmp/tmp.0J6yfGbozE/venv
$ python3 -m venv /tmp/tmp.0J6yfGbozE/venv
$ which pip
/tmp/tmp.0J6yfGbozE/venv/bin/pip

installing build deps into /tmp/tmp.0J6yfGbozE/venv
pip=/tmp/tmp.0J6yfGbozE/venv/bin/pip python3=/tmp/tmp.0J6yfGbozE/venv/bin/python3
$ pip install setuptools build wheel
Collecting setuptools
  Using cached setuptools-75.3.0-py3-none-any.whl.metadata (6.9 kB)
Collecting build
  Using cached build-1.2.2.post1-py3-none-any.whl.metadata (6.5 kB)
...
Successfully installed build-1.2.2.post1 packaging-24.1 pyproject_hooks-1.2.0 setuptools-75.3.0 wheel-0.44.0

$ rm -Rf dist/

building with /tmp/tmp.0J6yfGbozE/venv/bin/python3
$ /tmp/tmp.0J6yfGbozE/venv/bin/python3 setup.py bdist_wheel
...
adding 'testing/constraints-3.9.txt'
adding 'proto_plus-1.25.0.dist-info/LICENSE'
adding 'proto_plus-1.25.0.dist-info/METADATA'
adding 'proto_plus-1.25.0.dist-info/WHEEL'
adding 'proto_plus-1.25.0.dist-info/top_level.txt'
adding 'proto_plus-1.25.0.dist-info/RECORD'
removing build/bdist.linux-x86_64/wheel

built dist/proto_plus-1.25.0-py3-none-any.whl
-rw-rw-r-- 1 smoser smoser 230125 Nov  1 17:08 dist/proto_plus-1.25.0-py3-none-any.whl

...

FAIL: your wheel has docs/ testing/ build/
```

The second run will contain build/, docs/ and testing/ (from a dirty directory). The first will only contain docs and testing.

Here is a full list of the created wheel.
[unzip-wheel.txt](https://github.com/user-attachments/files/17604209/unzip-wheel.txt)

smoser:
https://setuptools.pypa.io/en/latest/userguide/package_discovery.html has a warning:

> Please have in mind that find_namespace: (setup.cfg),
> find_namespace_packages() (setup.py) and find (pyproject.toml)
> will scan all folders that you have in your project directory
> if you use a flat-layout.

That applies here.  Without this change, if you run: 'python3 setup.py bdist_wheel' then you may end up with build/ in your wheel, and in some cases even
docs/ and testing/ .

The fix is to only include proto and proto.*.

Fixes #503


smoser:
After a lot more testing and debugging I have a better idea of what is going on.
the presence of setuptools-scm in the wolfi build environment was what was ultimately causing 'docs/' and 'testing/' to get pulled into the wheel file.

I now think this is right, and is in-line with what is happening in other googleapis projects like https://github.com/googleapis/python-aiplatform/blob/f1da73b10d94d55a76d29de23fc3e198a37f78b6/setup.py#L36  and https://github.com/googleapis/python-pubsub/blob/5c78ac961199cd094ac4d29634fa4b6bfb73436a/setup.py#L62 .  Its better to be explicit and have to extend this list if there is ever a need.

I can modify the PR to be more like one of those if requested.

parthea (approved):

parthea:
Thanks @smoser!

PR #519, opened 2025-01-21
fix: construct messages with nested duration in protobuf 5.28+

### LINKED ISSUE #518: Constructing message from dictionary fails in protobuf 5.28.0 when message has a nested duration
As of protobuf 5.28.0, there is a bug in `proto-plus` where constructing a message from a dictionary fails when the message definition has a field of type `google.protobuf.duration_pb2.Duration`. Note this problem only exists when duration is in a nested message. See below:

```
    class Foo(proto.Message):
        another_field: duration_pb2.Duration = proto.Field(
            proto.MESSAGE,
            number=1,
            message=duration_pb2.Duration,
        )

    # this works
    Foo(another_field="300s")

    class Bar(proto.Message):
        some_field = proto.Field(proto.MESSAGE, number=1, message=Foo)

    # this raises `AttributeError: Fail to convert to Duration. Expected a timedelta like object got str: 'str' object has no attribute 'seconds'`
    bar = Bar({"some_field":{"another_field":"300s"}})
```

There is a change in the code path of message construction in proto-plus when using protobuf 5.28.0 and newer

In 5.27.5, we have
```
MessageMeta.__new__ → marshal.to_proto() → BaseMarshal.get_rule() → MessageRule.to_proto() → TypeError
```
where the exception is `TypeError: Message must be initialized with a dict: _default_package.Foo` raised at the location below`

In 5.28.0, we have 
```
MessageMeta.__new__ → marshal.to_proto() → BaseMarshal.get_rule() → MessageRule.to_proto() → AttributeError
```
where the exception is `AttributeError: Fail to convert to Duration. Expected a timedelta like object got str: 'str' object has no attribute 'seconds'`

Technically, this is not a breaking change in protobuf as they simply raised a different error. Proto-plus expected `TypeError` but `AttributeError` was raised instead.

See the re-production below:

In protobuf 5.27.5, run the code below
```
def test_duration_write_string_nested():
    class Foo(proto.Message):
        another_field: duration_pb2.Duration = proto.Field(
            proto.MESSAGE,
            number=1,
            message=duration_pb2.Duration,
        )

    class Bar(proto.Message):
        some_field = proto.Field(proto.MESSAGE, number=1, message=Foo)

    bar = Bar({"some_field":{"another_field":"300s"}})
    assert isinstance(bar.some_field.another_field, timedelta)
    assert isinstance(Bar.pb(bar).some_field.another_field, duration_pb2.Duration)
    assert bar.some_field.another_field.seconds == 300
    assert Bar.pb(bar).some_field.another_field.seconds== 300
```
You'll see `TypeError: Message must be initialized with a dict: _default_package.Foo` raised at the location below`

https://github.com/googleapis/proto-plus-python/blob/c54b31d71489785f20e5f140b8c43c7311fbfd97/proto/marshal/rules/message.py#L34-L37

In protobuf 5.27.5, the `TypeError` is caught, and then we successfully run the following code:
https://github.com/googleapis/proto-plus-python/blob/c54b31d71489785f20e5f140b8c43c7311fbfd97/proto/marshal/rules/message.py#L38-L45

In protobuf 5.28.0 and newer, you'll see `AttributeError` instead which is not caught in the code below and instead `AttributeError` is returned to the user

https://github.com/googleapis/proto-plus-python/blob/c54b31d71489785f20e5f140b8c43c7311fbfd97/proto/marshal/rules/message.py#L34-L45

### LINKED ISSUE #13350: google-cloud-secret-manager: duration "Fail to convert to Duration"
### Determine this is the right repository

- [X] I determined this is the correct repository in which to report this bug.

### Summary of the issue

**Context**

Trying to define a secret duration in "NNx" format (e.g. "300s") fails for latest versions of `protobuf` bundled with secretmanager. 

**Expected Behavior:**

(No error)

**Actual Behavior:**

```
AttributeError: Fail to convert to Duration. Expected a timedelta like object got str: 'str' object has no attribute 'seconds'
```

### API client name and version

google-cloud-secret-manager 2.21.1

### Reproduction steps: code

file: main.py
```python
import google.auth
import uuid
from google.cloud import secretmanager

_, project_id = google.auth.default()
client = secretmanager.SecretManagerServiceClient()

parent = f"projects/{project_id}"
ttl = "300s"

response = client.create_secret(
    request={
        "parent": f"projects/{project_id}",
        "secret_id": f"secret_{uuid.uuid4()}",
        "secret": {"replication": {"automatic": {}}, "ttl": "300s"},
    }
)
```

### Reproduction steps: supporting files

_No response_

### Reproduction steps: actual results

```
AttributeError: Fail to convert to Duration. Expected a timedelta like object got str: 'str' object has no attribute 'seconds'
```

### Reproduction steps: expected results

(Successful creation)

### OS & version + platform

MacOS 15.1.1 (24B91)

### Python environment

Python 3.13.0

### Python dependencies

cachetools==5.5.0
certifi==2024.8.30
charset-normalizer==3.4.0
google-api-core==2.24.0
google-auth==2.37.0
google-cloud-secret-manager==2.21.1
googleapis-common-protos==1.66.0
grpc-google-iam-v1==0.13.1
grpcio==1.68.1
grpcio-status==1.68.1
idna==3.10
proto-plus==1.25.0
protobuf==5.29.1
pyasn1==0.6.1
pyasn1_modules==0.4.1
requests==2.32.3
rsa==4.9
urllib3==2.2.3

### Additional context

Dependencies based on a current `pip install google-cloud-secret-manager`, which comes with `protobuf==5.29.1`. 

Manually running `pip install protobuf==5.27.5`, **the error does not occur**. 

Error exists from `protobuf==5.28.0` onwards, without any changes to any other dependencies.

parthea:
This fix is similar to https://github.com/googleapis/proto-plus-python/pull/479 where message construction fails when a specific type, in this case `google.protobuf.duration_pb2.Duration`, is nested.

Fixes https://github.com/googleapis/proto-plus-python/issues/518
Fixes https://github.com/googleapis/google-cloud-python/issues/13350

### DIFF
--- a/proto/marshal/rules/message.py
+++ b/proto/marshal/rules/message.py
@@ -41,8 +41,8 @@ class MessageRule:
                 # - an int64/string issue.
                 # - a missing key issue in case a key only exists with a `_` suffix.
                 #   See related issue: https://github.com/googleapis/python-api-core/issues/227.
-                # - a missing key issue due to nested struct. See: b/321905145.
-                # - a missing key issue due to nested duration. See: b/383651746.
+                # - a missing key issue due to nested struct. See: https://github.com/googleapis/proto-plus-python/issues/424.
+                # - a missing key issue due to nested duration. See: https://github.com/googleapis/google-cloud-python/issues/13350.
                 return self._wrapper(value)._pb
         return value
 
--- a/proto/marshal/rules/message.py
+++ b/proto/marshal/rules/message.py
@@ -34,14 +34,15 @@ class MessageRule:
             try:
                 # Try the fast path first.
                 return self._descriptor(**value)
-            except (TypeError, ValueError) as ex:
-                # If we have a TypeError or Valueerror,
+            except (TypeError, ValueError, AttributeError) as ex:
+                # If we have a TypeError, Valueerror or AttributeError,
                 # try the slow path in case the error
                 # was:
                 # - an int64/string issue.
                 # - a missing key issue in case a key only exists with a `_` suffix.
                 #   See related issue: https://github.com/googleapis/python-api-core/issues/227.
                 # - a missing key issue due to nested struct. See: b/321905145.
+                # - a missing key issue due to nested duration. See: b/383651746.
                 return self._wrapper(value)._pb
         return value
 
--- a/proto/marshal/rules/message.py
+++ b/proto/marshal/rules/message.py
@@ -35,7 +35,7 @@ class MessageRule:
                 # Try the fast path first.
                 return self._descriptor(**value)
             except (TypeError, ValueError, AttributeError) as ex:
-                # If we have a TypeError, Valueerror or AttributeError,
+                # If we have a TypeError, ValueError or AttributeError,
                 # try the slow path in case the error
                 # was:
                 # - an int64/string issue.
--- a/tests/test_marshal_types_dates.py
+++ b/tests/test_marshal_types_dates.py
@@ -241,6 +241,26 @@ def test_duration_write_string():
     assert Foo.pb(foo).ttl.seconds == 120
 
 
+def test_duration_write_string_nested():
+    class Foo(proto.Message):
+        another_field: duration_pb2.Duration = proto.Field(
+            proto.MESSAGE,
+            number=1,
+            message=duration_pb2.Duration,
+        )
+
+    Foo(another_field="300s")
+
+    class Bar(proto.Message):
+        some_field = proto.Field(proto.MESSAGE, number=1, message=Foo)
+
+    bar = Bar({"some_field": {"another_field": "300s"}})
+    assert isinstance(bar.some_field.another_field, timedelta)
+    assert isinstance(Bar.pb(bar).some_field.another_field, duration_pb2.Duration)
+    assert bar.some_field.another_field.seconds == 300
+    assert Bar.pb(bar).some_field.another_field.seconds == 300
+
+
 def test_duration_del():
     class Foo(proto.Message):
         ttl = proto.Field(
--- a/tests/test_marshal_types_dates.py
+++ b/tests/test_marshal_types_dates.py
@@ -243,22 +243,22 @@ def test_duration_write_string():
 
 def test_duration_write_string_nested():
     class Foo(proto.Message):
-        another_field: duration_pb2.Duration = proto.Field(
+        foo_field: duration_pb2.Duration = proto.Field(
             proto.MESSAGE,
             number=1,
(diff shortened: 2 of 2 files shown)


proto/marshal/rules/message.py
```diff
@@ -34,14 +34,15 @@ def to_proto(self, value):
...
-                # If we have a TypeError or Valueerror,
+            except (TypeError, ValueError, AttributeError) as ex:
+                # If we have a TypeError, Valueerror or AttributeError,
                 # try the slow path in case the error
                 # was:
                 # - an int64/string issue.
                 # - a missing key issue in case a key only exists with a `_` suffix.
                 #   See related issue: https://github.com/googleapis/python-api-core/issues/227.
                 # - a missing key issue due to nested struct. See: b/321905145.
+                # - a missing key issue due to nested duration. See: b/383651746.
```
vchudnov-g, line 45: Use external links as much as possible.

```suggestion
                # - a missing key issue due to nested struct. See: https://github.com/googleapis/proto-plus-python/issues/424.
                # - a missing key issue due to nested duration. See: https://github.com/googleapis/google-cloud-python/issues/13350.
```

proto/marshal/rules/message.py
```diff
@@ -34,14 +34,15 @@ def to_proto(self, value):
             try:
                 # Try the fast path first.
                 return self._descriptor(**value)
-            except (TypeError, ValueError) as ex:
-                # If we have a TypeError or Valueerror,
+            except (TypeError, ValueError, AttributeError) as ex:
+                # If we have a TypeError, Valueerror or AttributeError,
```
vchudnov-g, line 38:
```suggestion
                # If we have a TypeError, ValueError or AttributeError,
```

tests/test_marshal_types_dates.py
```diff
@@ -241,6 +241,26 @@ class Foo(proto.Message):
     assert Foo.pb(foo).ttl.seconds == 120
 
 
+def test_duration_write_string_nested():
```
vchudnov-g, line 244: nit: for clarity, I suggest:
s/`another_field`/`foo_field`/g
s/`some_field`/`bar_field`/g

vchudnov-g (approved):
Thanks for fixing this! Only minor comments.

tests/test_marshal_types_dates.py
parthea: Fixed in fcb5bb26557c714fd14724fa541bdfc15219a958

PR #559, opened 2026-01-24
fix: remove float_precision for protobuf 7

### LINKED ISSUE #15099: Deprecate `float_precision` argument
As per the following stack trace, `float_precision` is deprecated in protobuf. We should emit deprecation warnings when the `float_precision` argument is set :

https://github.com/googleapis/proto-plus-python/blob/8ecf01302126b4950170d267e71875c82b5a8f89/proto/message.py#L469

https://github.com/googleapis/proto-plus-python/blob/8ecf01302126b4950170d267e71875c82b5a8f89/proto/message.py#L559

This deprecation started in protobuf 6.32.x
https://github.com/protocolbuffers/protobuf/releases/tag/v32.0

```
__________________________________________________________________________________________________________________________ test_serialize_to_dict_float_precision __________________________________________________________________________________________________________________________

    def test_serialize_to_dict_float_precision():
        class Squid(proto.Message):
            mass_kg = proto.Field(proto.FLOAT, number=1)
    
        s = Squid(mass_kg=3.14159265)
    
>       s_dict = Squid.to_dict(s, float_precision=3)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

tests/test_message.py:336: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
proto/message.py:610: in to_dict
    return MessageToDict(
.nox/unit-3-13-implementation-upb/lib/python3.13/site-packages/google/protobuf/json_format.py:158: in MessageToDict
    printer = _Printer(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <google.protobuf.json_format._Printer object at 0x7fb6d1f15940>, preserving_proto_field_name = True, use_integers_for_enums = True, descriptor_pool = None, float_precision = 3, always_print_fields_with_no_presence = True

    def __init__(
        self,
        preserving_proto_field_name=False,
        use_integers_for_enums=False,
        descriptor_pool=None,
        float_precision=None,
        always_print_fields_with_no_presence=False,
    ):
      self.always_print_fields_with_no_presence = (
          always_print_fields_with_no_presence
      )
      self.preserving_proto_field_name = preserving_proto_field_name
      self.use_integers_for_enums = use_integers_for_enums
      self.descriptor_pool = descriptor_pool
      if float_precision:
>       warnings.warn(
            'float_precision option is deprecated for json_format. '
            'This will turn into error in 7.34.0, please remove it '
            'before that.'
        )
E       UserWarning: float_precision option is deprecated for json_format. This will turn into error in 7.34.0, please remove it before that.

.nox/unit-3-13-implementation-upb/lib/python3.13/site-packages/google/protobuf/json_format.py:195: UserWarning
```

daniel-sanche:
Fixes https://github.com/googleapis/google-cloud-python/issues/15099

Protobuf 7 is removing the float_prevision argument. This PR ignores it when passed in 7.x+, and raises a DeprecationWarning

### DIFF
--- a/proto/message.py
+++ b/proto/message.py
@@ -946,11 +946,12 @@ def _message_to_map(
         # float_precision removed in protobuf 7
         if _PROTOBUF_MAJOR_VERSION in ("3", "4", "5", "6"):
             kwargs["float_precision"] = float_precision
-            warning_msg = "`float_precision` will be removed in Protobuf 7.x."
-
         else:  # pragma: NO COVER
-            warning_msg = "`float_precision` was removed in Protobuf 7.x+, and will be ignored."
-        warnings.warn(warning_msg, DeprecationWarning, stacklevel=3)
+            warnings.warn(
+                "`float_precision` was removed from Protobuf 7.x, and will be ignored",
+                DeprecationWarning,
+                stacklevel=3,
+            )
 
     return map_fn(cls.pb(instance), **kwargs)
 
--- a/proto/message.py
+++ b/proto/message.py
@@ -457,6 +457,46 @@ class MessageMeta(type):
             or including_default_value_fields
         )
 
+    @staticmethod
+    def _to_map(
+        cls,
+        map_fn,
+        instance,
+        *,
+        including_default_value_fields=None,
+        float_precision=None,
+        always_print_fields_with_no_presence=None,
+        **kwargs,
+    ):
+        """
+        Helper for logic for to_dict and to_json
+        """
+        print_fields = cls._normalize_print_fields_without_presence(
+            always_print_fields_with_no_presence, including_default_value_fields
+        )
+
+        # The `including_default_value_fields` argument was removed from protobuf 5.x
+        # and replaced with `always_print_fields_with_no_presence` which very similar but has
+        # handles optional fields consistently by not affecting them.
+        # The old flag accidentally had inconsistent behavior between proto2
+        # optional and proto3 optional fields.
+        if PROTOBUF_VERSION[0] in ("3", "4"):
+            kwargs["including_default_value_fields"] = print_fields
+        else:
+            kwargs["always_print_fields_with_no_presence"] = print_fields
+
+        if float_precision:
+            # float_precision removed in protobuf 7
+            if int(PROTOBUF_VERSION[0]) < 7:
+                kwargs["float_precision"] = float_precision
+            else:
+                warnings.warn(
+                    "The argument `float_precision` has been removed from Protobuf 7.x.",
+                    DeprecationWarning,
+                )
+
+        return map_fn(cls.pb(instance), **kwargs)
+
     def to_json(
         cls,
         instance,
@@ -491,7 +531,7 @@ class MessageMeta(type):
                 An indent level of 0 or negative will only insert newlines.
                 Pass None for the most compact representation without newlines.
             float_precision (Optional(int)): If set, use this to specify float field valid digits.
-                Default is None.
+                Default is None. [DEPRECATED] float_precision was removed in Protobuf 7.x.
             always_print_fields_with_no_presence (Optional(bool)): If True, fields without
                 presence (implicit presence scalars, repeated fields, and map fields) will
                 always be serialized. Any field that supports presence is not affected by
@@ -501,36 +541,7 @@ class MessageMeta(type):
         Returns:
             str: The json string representation of the protocol buffer.
         """
-
-        print_fields = cls._normalize_print_fields_without_presence(
-            always_print_fields_with_no_presence, including_default_value_fields
-        )
-
-        if PROTOBUF_VERSION[0] in ("3", "4"):
-            return MessageToJson(
-                cls.pb(instance),
-                use_integers_for_enums=use_integers_for_enums,
-                including_default_value_fields=print_fields,
-                preserving_proto_field_name=preserving_proto_field_name,
-                sort_keys=sort_keys,
-                indent=indent,
-                float_precision=float_precision,
-            )
-        else:
-            # The `including_default_value_fields` argument was removed from protobuf 5.x
-            # and replaced with `always_print_fields_with_no_presence` which very similar but has
-            # handles optional fields consistently by not affecting them.
-            # The old flag accidentally had inconsistent behavior between proto2
-            # optional and proto3 optional fields.
-            return MessageToJson(
-                cls.pb(instance),
-                use_integers_for_enums=use_integers_for_enums,
-                always_print_fields_with_no_presence=print_fields,
-                preserving_proto_field_name=preserving_proto_field_name,
-                sort_keys=sort_keys,
-                indent=indent,
-                float_precision=float_precision,
-            )
+        return cls._to_map(map_fn=MessageToJson, **locals())
 
     def from_json(cls, payload, *, ignore_unknown_fields=False) -> "Message":
         """Given a json string representing an instance,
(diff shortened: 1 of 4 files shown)


proto/message.py
```diff
@@ -36,6 +36,10 @@
 
 PROTOBUF_VERSION = google.protobuf.__version__
 
+# extract the major version code
+_separator_idx = PROTOBUF_VERSION.find('.')
+_PROTOBUF_MAJOR_VERSION = PROTOBUF_VERSION[:_separator_idx] if _separator_idx != -1 else PROTOBUF_VERSION
```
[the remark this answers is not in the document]
daniel-sanche: This was done for performance reasons, but after running timeit, it didn't seem impactful. Changed

proto/message.py
```diff
@@ -966,4 +915,45 @@ def pb(self) -> Type[message.Message]:
...
+    if float_precision:
+        # float_precision removed in protobuf 7
+        if _PROTOBUF_MAJOR_VERSION in ("3", "4", "5", "6"):
+            kwargs["float_precision"] = float_precision
+        else:  # pragma: NO COVER
+            warnings.warn(
+                "The argument `float_precision` has been removed from Protobuf 7.x.",
+                DeprecationWarning,
+                stacklevel=3,
+            )
```
parthea: Should we emit this warning regardless of the protobuf version to give maintainers a heads up about this deprecation/removal?

parthea: Can we emit a different warning if the `float_precision` argument was provided with protobuf 7.x, to say that it's been ignored?

daniel-sanche: > Can we emit a different warning if the float_precision argument was provided with protobuf 7.x, to say that it's been ignored?

I made the change to say explictly that it's being ignored in 7.x

> Should we emit this warning regardless of the protobuf version to give maintainers a heads up about this deprecation/removal?

For this one, I started making the change, but realized we are also getting a duplicate warning bubble up from the proto library, so I think we can rely on that one instead. I removed it from the ignored warnings list. Let me know what you think

parthea:
@daniel-sanche , Please could you address the failing tests?

proto/message.py
parthea: We only get the duplicate warning with protobuf 6.x . Can we emit the warning for protobuf 3, 4, 5 or do we consider those soon to be unsupported?

daniel-sanche: I made a change to suppress the protobuf warning, and emit our own. 

I think we could also get away with just letting the protobuf library handle the pre-7.x warnings, but this way is more consistent. Let me know what you think

proto/message.py
```diff
@@ -966,4 +915,48 @@ def pb(self) -> Type[message.Message]:
...
+):
+    """
+    Helper for logic for Message.to_dict and Message.to_json
+    """
+
+    # The `including_default_value_fields` argument was removed from protobuf 5.x
+    # and replaced with `always_print_fields_with_no_presence` which very similar but has
+    # handles optional fields consistently by not affecting them.
+    # The old flag accidentally had inconsistent behavior between proto2
+    # optional and proto3 optional fields.
```
chalmerlowe: Updated incorrect grammar and tweaked the wording of the closing phrase.

```suggestion
    # The `including_default_value_fields` argument was removed from protobuf 5.x
    # and replaced with `always_print_fields_with_no_presence` which is similar but
    # handles optional fields consistently by not affecting them.
    # The old flag accidentally had inconsistent behavior between optional fields
    # in proto2 and proto3.
```

proto/message.py
```diff
@@ -966,4 +915,48 @@ def pb(self) -> Type[message.Message]:
...
+        if _PROTOBUF_MAJOR_VERSION in ("3", "4", "5", "6"):
+            kwargs["float_precision"] = float_precision
+            warning_msg = "`float_precision` will be removed in Protobuf 7.x."
+
+        else:  # pragma: NO COVER
+            warning_msg = (
+                "`float_precision` was removed in Protobuf 7.x+, and will be ignored."
+            )
+        warnings.warn(warning_msg, DeprecationWarning, stacklevel=3)
+    # supress similar float_precision warning from protobuf library
```
chalmerlowe: Updated spelling:

```suggestion
    # suppress similar float_precision warning from protobuf library.
```

tests/test_json.py
```diff
@@ -253,13 +253,36 @@ class Squid(proto.Message):
     assert re.search(r"massKg.*name", j)
 
 
-# TODO: https://github.com/googleapis/proto-plus-python/issues/390
 def test_json_float_precision():
+    if int(proto.message._PROTOBUF_MAJOR_VERSION) >= 7:
```
chalmerlowe: NO ACTION:

I situations like this, I often choose to decorate a test with `@pytest.mark.parametrize()` and include one or more happy paths and sad paths (labeled with IDs for clarity when looking at the pytest results) rather than writing two separate tests. Avoids a lot of duplicate code. Not saying you need to do anything here. Just pointing it out as an option for future tests down the road.

chalmerlowe (approved):
Approved with suggested fixes for minor spelling and grammar errors.

Also, includes a suggestion for future work, that is not required to be incorporated here (NO ACTION).

proto/message.py
```diff
@@ -966,4 +915,48 @@ def pb(self) -> Type[message.Message]:
...
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=UserWarning)
```
parthea: One downside of this approach is that we will no longer see new user warnings. For example, in October 2025, https://pypi.org/project/protobuf/6.33.0/ included a new deprecation warning for the `float_precision` argument, within the same major version. See the release notes for https://github.com/protocolbuffers/protobuf/releases/tag/v33.0-rc1

Can we use `warnings.filterwarnings` to narrow down the warnings that are filtered to the one that includes the text `float_precision`?

```
warnings.filterwarnings("ignore", message=".*float_precision.*")
```

tests/test_json.py
daniel-sanche: Good point. I think they started out different, but the code is pretty similar now. I'll take another look

proto/message.py
daniel-sanche: good point, fixed

pytest.ini
```diff
@@ -8,4 +8,3 @@ filterwarnings =
     ignore:.*The argument `including_default_value_fields` has been removed.*:DeprecationWarning
     # Remove once deprecated field `float_precision` is removed
     # See https://github.com/googleapis/proto-plus-python/issues/547
```
parthea:
```suggestion
```

parthea (approved):

parthea (approved):

